PR #8, opened 2021-04-06
Payload Receiver Service (Fluentd->FastAPI) v0 

sanjay920:


src/payload-receiver-service/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3.8
```
tybalex, line 1: a suggestion that apply to all of our images:
for images that the usage is as lightweight as this one, we should probably use python:3.8-slim instead.

tybalex (approved):
look good to me! just a minor suggestion as commented.
Let's wait for the PR #7 to be merged , and we can then merge this ! :)

AmartC (approved):
Looks good to me! We can merge this after the other PR is finalized.

src/payload-receiver-service/app/main.py
```diff
@@ -0,0 +1,82 @@
...
+from nats_wrapper import NatsWrapper
+
+app = FastAPI()
+logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(name)s - %(message)s")
+nw = None
+
+
+@app.on_event("startup")
+async def startup_event():
+    global nw
```
tybalex, line 20: one more minor comments is, nw is already a global variable so we don't need to do `global nw` here.

src/payload-receiver-service/Dockerfile
sanjay920: using `python:3.8-slim ` now

src/payload-receiver-service/app/main.py
sanjay920: `global nw` needs to be here because `@app.on_event("startup")` is a FastAPI related processes 

It can be removed from `get_nats()` though.

PR #11, opened 2021-04-09
logs preprocessing service

sanjay920:


.pre-commit-config.yaml
```diff
@@ -2,22 +2,22 @@
...
       - id: check-yaml
       - id: check-added-large-files
   - repo: https://github.com/PyCQA/isort
-    rev: 5.5.1
+    rev: 5.8.0
     hooks:
       - id: isort
   - repo: https://github.com/humitos/mirrors-autoflake.git
-    rev: v1.3
+    rev: v1.1
```
tybalex, line 15: this is weird why updating pre-commit would revert the version of autoflake to an older version?

src/preprocessing-service/masker.py
```diff
@@ -0,0 +1,126 @@
...
+        self.masker = RegexMasker(
+            masking_instructions,
+            masking_instructions_before_value_assigning_token_split,
+        )
+
+    def mask(self, content: str):
+        if self.masker is not None:
+            return self.masker.mask(content)
+        else:
+            return content
```
tybalex, line 126: in this case when there's no masker, should it raise an exception?

src/preprocessing-service/preprocess.py
```diff
@@ -0,0 +1,95 @@
...
+# Third Party
+import pandas as pd
+from elasticsearch import AsyncElasticsearch
+from elasticsearch.helpers import async_streaming_bulk
+from masker import LogMasker
+from nats_wrapper import NatsWrapper
+
+logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(name)s - %(message)s")
+
+ES_ENDPOINT = "https://opendistro-es-client-service.default.svc.cluster.local:9200"
```
tybalex, line 14: let's make this an ENV variable? so it can be specified in the yaml file.

tybalex (changes requested):
requested a few minor changes.

src/preprocessing-service/preprocess.py
AmartC: Yes, agreed add to the yaml file.

src/preprocessing-service/preprocess.py
```diff
@@ -0,0 +1,95 @@
...
+
+es = AsyncElasticsearch(
+    [ES_ENDPOINT],
+    port=9200,
```
AmartC, line 18: please add the port as an ENV variable in the yaml file as well.

AmartC (changes requested):

AmartC:
I have also noticed that the preprocessing service still gets restarted a good number of times. Is this expected right now?

sanjay920:
> I have also noticed that the preprocessing service still gets restarted a good number of times. Is this expected right now?

Service will **not** fail & restart if you use payload-receiver-service from v0.0
Make sure you are running the latest code

src/preprocessing-service/masker.py
sanjay920: you're right this exception will never be hit because `self.masker` is guaranteed to exist and be initialized

src/preprocessing-service/preprocess.py
sanjay920: added

src/preprocessing-service/preprocess.py
sanjay920: port not necessary in yaml

.pre-commit-config.yaml
sanjay920: not sure why. it had me update pre-commit before checks could be run. doesn't seem to be an issue for now

tybalex (approved):

PR #15, opened 2021-04-20
Update payload-receiver-service to route control plane logs to different NATs subject.

AmartC:
payload-receiver-service has been updated to route control plane logs to Nats subject raw_control_logs.

### DIFF
--- a/src/preprocessing-service/preprocess.py
+++ b/src/preprocessing-service/preprocess.py
@@ -35,7 +35,7 @@ async def doc_generator(df):
     df["nulog_confidence"] = -1.0
     df["drain_matched_template_id"] = -1.0
     df["drain_matched_template_support"] = -1.0
-    df["anomaly_level"] = "info"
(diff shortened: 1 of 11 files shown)


src/preprocessing-service/preprocess.py
```diff
@@ -73,7 +73,24 @@ async def mask_logs(nw, loop, queue):
         for index, row in payload_data_df.iterrows():
             masked_logs.append(masker.mask(row["log"]))
         payload_data_df["masked_log"] = masked_logs
-        await nw.publish("preprocessed_logs", payload_data_df.to_json().encode())
+        if "filename" in payload_data_df.columns:
```
sanjay920, line 76: payload can contain both control plane and application logs. this approach assumes it's one or the other

sanjay920 (changes requested):

sanjay920 (approved):

PR #16, opened 2021-04-20
nulog model bug fix.

tybalex:
This PR fixes 2 bugs:
1. a corner case that predictions results not calculated correctly
2. handle the case that input message is empty string.


sanjay920 (approved):

AmartC (commented):
Looks good to me!

PR #17, opened 2021-04-21
Training controller load management

AmartC:
Update training-controller service so if it is given numerous time intervals from DRAIN, based on the total number of logs that can be fetched based on memory constraints,  for each time interval, it will fetch the total number of logs within the time interval and use that to proportionally determine how many logs to fetch from each time interval through Elasticdump.


AmartC:
training-controller-load-management branch has been merged with training-controller branch so this PR can now be closed.

PR #18, opened 2021-04-23
klog timestamp parsing

sanjay920:
Add log timestamp parsing to preprocessing service

### DIFF
--- a/src/preprocessing-service/preprocess.py
+++ b/src/preprocessing-service/preprocess.py
@@ -35,7 +35,7 @@ async def doc_generator(df):
     df["nulog_confidence"] = -1.0
     df["drain_matched_template_id"] = -1.0
     df["drain_matched_template_support"] = -1.0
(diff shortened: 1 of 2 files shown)


src/preprocessing-service/preprocess.py
```diff
@@ -35,7 +35,7 @@ async def doc_generator(df):
     df["nulog_confidence"] = -1.0
     df["drain_matched_template_id"] = -1.0
     df["drain_matched_template_support"] = -1.0
-    df["anomaly_level"] = "info"
+    df["anomaly_level"] = "normal"
```
tybalex, line 38: `Normal` or `normal`?

sanjay920: nulog inferencing is capitalized so will capitalize it here

tybalex (approved):
look good.

PR #25, opened 2021-04-29
Control plane nulog inference service

tybalex:


sanjay920 (approved):

src/nulog-inference-service/build.sh
```diff
@@ -1,4 +1,6 @@
+#IMAGE_NAME=tybalex/nulog-inference-service-control-plane:v0.0
```
AmartC, line 1: So the same build.sh file is used to build the nulog-inference-service-control-plane and nulog-inference-service images?

AmartC (approved):
Looks good overall. Just the question about build.sh which is a clarification.

src/nulog-inference-service/build.sh
tybalex: yes, the codes are same. Only difference is, in `nulog-CP` image, we also include a pre-train CP model.

PR #26, opened 2021-04-29
Add opnictl tool

galal-hussein:
Add opnictl code to the opni repo


tybalex (approved):
I only reviewed the opnictl files, and they look good! I assume the go vendering files are fine.

one question I have is, will we have a file such as `values.yaml` basically contains any values of parameters used in other yaml files? For example, parameters such as `%ES_USER`, `%ES_PASSWORD`.
@galal-hussein

galal-hussein:
> I only reviewed the opnictl files, and they look good! I assume the go vendering files are fine.
> 
> one question I have is, will we have a file such as `values.yaml` basically contains any values of parameters used in other yaml files? For example, parameters such as `%ES_USER`, `%ES_PASSWORD`.
> @galal-hussein

no this is strictly is used within the opnictl tool, and will be utilized as flags that can be passed to the tool

tybalex:
> > I only reviewed the opnictl files, and they look good! I assume the go vendering files are fine.
> > one question I have is, will we have a file such as `values.yaml` basically contains any values of parameters used in other yaml files? For example, parameters such as `%ES_USER`, `%ES_PASSWORD`.
> > @galal-hussein
> 
> no this is strictly is used within the opnictl tool, and will be utilized as flags that can be passed to the tool

OK. I think in the future we should also make those hard-coded urls of nats and Elasticsearch (`NATS_URL` and `ES_ENDPOINT`) as variables, are you going to make these as flags with default values as well?

AmartC (changes requested):
For Nats, please use nats-client.kube-system.svc. Currently,leading to errors.

AmartC (changes requested):
Looks good for most part. However, please do update the documentation for what commands need to be run if using MacBook.

src/dashboard/Kibana/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM nikolaik/python-nodejs:python3.8-nodejs14-slim
```
tybalex, line 1: not a must-do but it only needs python, so let's use `FROM python:3.8-slim`.

PR #28, opened 2021-05-06
Address bug in training controller so model training is launched from Kibana.

AmartC:
Address a bug to give the user the option to launch the training of a Nulog model from Kibana.


tybalex (approved):
look good. Just 1 redundant line.

PR #32, opened 2021-05-07
Nulog speedup

tybalex:
Major update:
Speed up Nulog inference by using a dict to memorize history predictions.

Minor update:
fix elasticsearch update issue. - (failed to compile inline script)


src/nulog-inference-service/start-nulog-inference.py
```diff
@@ -94,38 +98,29 @@ async def doc_generator(df):
...
-
             df.rename(columns={"log_id": "_id"}, inplace=True)
-            df["script"] = (
-                df["script"]
-                + "ctx._source.nulog_confidence = "
-                + df["nulog_confidence"].map(str)
-                + ";"
-            )
-
+            df["script"] = [
```
tybalex, line 105: @sanjay920 FYI, this is where I fixed the inline script. 
basically the idea is - `script` has three fields: 
1. `source` for your script, but it should be a fix string rather than different strings in each call, such as `df["nulog_confidence"].map(str)`, then ES only needs to compile the source script once instead of many times
2. "lang" : "painless".
3. "params": the parameters that the value always changes, in my cases is the nulog_score value.

AmartC (approved):

PR #39, opened 2021-05-11
merge 2 nulog images into 1

tybalex:
the nulog workload image and control-plane image now get merged to use the same image
This makes the maintenance easier.
the nulog-CP-service will read from `control-plane-output/` dir with the pre-trained model, while the regular nulog-service (for workload apps) will read from `output/` dir as usual.


AmartC (approved):
Looks good to me.

PR #40, opened 2021-05-12
Drain service init

sanjay920:


src/drain-service/drain_training_inferencing.py
```diff
@@ -0,0 +1,345 @@
...
+        weighted_mean, weighted_std = weighted_avg_and_std(num_clusters, weights)
+        weighted_vol = weighted_std / np.mean(num_clusters[:10])
+        logging.info(
+            "ITERATION {}: vol= {} weighted_vol= {} normal_periods={}".format(
+                iteration, vol, weighted_vol, str(normal_periods)
+            )
+        )
+
+        if len(num_total_clusters_tracking_queue) > 3:
+            if weighted_vol >= 0.199:
```
AmartC, line 266: Fine for now but may be good in the future to give the user an option to set this variable as well if they would like to do so.

AmartC (approved):
Looks good. Just that one comment I have but for now it looks fine.

src/drain-service/requirements.txt
```diff
@@ -0,0 +1,8 @@
+jsonpickle==1.5.1
+kafka-python==2.0.1
```
tybalex, line 2: do we need Kafka-python?

tybalex (approved):

src/drain-service/drain_training_inferencing.py
```diff
@@ -0,0 +1,345 @@
...
+        anomalies_data = msg.data.decode()
+        logging.info("got anomaly payload")
+        await logs_to_update_es.put(pd.read_json(anomalies_data, dtype={"_id": object}))
+
+    await nw.subscribe(
+        nats_subject="preprocessed_logs",
+        payload_queue=None,
+        subscribe_handler=subscribe_handler,
+    )
+    await nw.subscribe(
```
tybalex, line 58: would it be better to have a drain pod handling workload logs and another handling CP logs? If the load is heavy.

src/drain-service/drain_training_inferencing.py
sanjay920: What specifically? The minimum number of time periods (3 20-second periods right now) elapsed before checking volatility? Or the 0.199 value?
can you name an example in which the user would want to set this value? 
I was thinking it would be better to let them select a low/medium/high sensitivity rather than tweak this value explicitly.

src/drain-service/drain_training_inferencing.py
sanjay920: Same pod is fine. I do think we should create a different DRAIN model for CP. Added this as a TODO

src/drain-service/requirements.txt
sanjay920: removed kafka-python and redis

PR #52, opened 2021-05-20
Nats reconnect, ES connection stability improvements

sanjay920:


src/drain-service/drain_training_inferencing.py
```diff
@@ -43,31 +50,24 @@ async def anomalies_subscription_handler(msg):
...
+        payload_queue=None,
+        subscribe_handler=subscribe_handler,
+    )
+    await nw.subscribe(
+        nats_subject="preprocessed_logs_control_plane",
+        payload_queue=None,
+        subscribe_handler=subscribe_handler,
+    )
+    await nw.subscribe(
+        nats_subject="anomalies",
```
tybalex, line 64: what does this subject do?

src/preprocessing-service/preprocess.py
```diff
@@ -153,13 +163,21 @@ async def es_log_management():
         await asyncio.sleep(3600)  # deletion once an hour
 
 
+async def init_nats():
```
tybalex, line 166: I noticed that this function appears everywhere, is there a better way to do this?

src/training_controller/training_controller.py
```diff
@@ -139,16 +138,15 @@ async def es_training_signal_coroutine(signals_queue: asyncio.Queue):
 def job_not_currently_running(job_name, namespace=NAMESPACE):
     try:
         jobs = api_instance.list_namespaced_job(namespace, timeout_seconds=60)
+        for job in jobs.items:
+            if job.metadata.name == job_name:
+                return False
+        return True
```
tybalex, line 144: should this function return False if it drops into exception condition?

src/utils/nats_wrapper.py
```diff
@@ -50,16 +45,16 @@ async def reconnected_cb():
             "reconnected_cb": reconnected_cb,
             "disconnected_cb": on_disconnect,
             "servers": [self.NATS_SERVER_URL],
+            "max_reconnect_attempts": -1,
```
tybalex, line 48: so no reconnect attempts?

AmartC: I think the -1 indicates that it will try reconnecting indefinitely

AmartC (approved):
Tested out the changes and it's been running stably.

src/drain-service/drain_training_inferencing.py
sanjay920: it is where anomalies marked by drain go to. this is set up as a NATS queue instead of an Asyncio queue because it will allow us to have an ES updating service in the future. for now we don't see any benefit from it

src/preprocessing-service/preprocess.py
sanjay920: unfortunately we need to initialize nats before passing its instance off to the other coroutines. let me know if you have any suggestions

src/training_controller/training_controller.py
sanjay920: it doesn't need to return anything if it falls into an exception.
I refactored because this block of code was being executed outside of the try-except block, which could lead to a scenario in which `jobs` will not exist

src/utils/nats_wrapper.py
sanjay920: indefinitely

src/drain-service/drain_training_inferencing.py
tybalex: ok

src/preprocessing-service/preprocess.py
tybalex: it's fine, let's not worry about this for now then.

tybalex (approved):

PR #67, opened 2021-06-23
Installation options

kralicky:
- Removed Traefik and Local Path Provisioner from all installations
- ~~Adjusted demo API to add support for existing rancher logging installation~~ No change needed
- Removed `--quickstart` flag from `opnictl create` and added flags with more explicit functions
- Added interactive prompts for a few missing command line arguments that don't have easily inferable defaults
- Fixes/improvements to Tiltfile
- Switch kubebuilder project to use multi-group mode
- Some minor adjustments and fixes to status condition text


controllers/demo/opnidemo_controller.go
```diff
@@ -161,18 +164,33 @@ func (r *OpniDemoReconciler) reconcileInfraStack(
...
 			return ctrl.Result{Requeue: true}, r.Create(ctx, object)
 		} else if err != nil {
 			return ctrl.Result{}, err
+		} else {
+			if ds, ok := object.(*appsv1.DaemonSet); ok {
+				if ds.Status.NumberReady != ds.Status.DesiredNumberScheduled {
+					opniDemo.Status.Conditions = append(opniDemo.Status.Conditions,
+						fmt.Sprintf("Waiting for daemonset %s to become ready", ds.Name))
+				}
+			}
```
dbason, line 193: We only check for the DaemonSets?

kralicky: This is just checking the nvidia daemonset and helm controller resources, I think the only workload there worth checking the status of is the nvidia daemonset, unless you have any other suggestions

dbason (approved):

PR #74, opened 2021-07-02
Add telemetry to opni manager

dbason:
Adds the upgrade checker/telemetry in to the Opni controller manager.  Also minor tidyups for Tilt.


main.go
```diff
@@ -42,6 +44,11 @@ import (
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
+	Version  = "v0.1.0-dev"
```
kralicky, line 47: should this just be "dev" do you think? this is the version that would show up if `go get` was used to download the binary.

dbason: Sure thing, I initially had it this way to test the semver comparison.

kralicky: should we disable update checking if the verison is dev?

main.go
```diff
@@ -42,6 +44,11 @@ import (
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
+	Version  = "dev"
```
tybalex, line 47: this would be the release version number , right? eg.. `v0.1.3`? do we have to manually change this value for every release?

dbason: Nope we set it using ldflag when we build

scripts/manager
```diff
@@ -1,5 +1,9 @@
...
-cd $(dirname $0)/..
+cd $(dirname $0)
 
-CGO_ENABLED=0 go build -ldflags '-w -s' -o bin/manager .
\ No newline at end of file
+source version
+
+cd ..
+
+CGO_ENABLED=0 go build -ldflags "-w -s -X main.Version=$VERSION" -o bin/manager .
```
dbason, line 9: @tybalex this line here is where we set the version at build time

kralicky (approved):

PR #84, opened 2021-07-27
Update README with installation instructions

sanjay920:


README.md
```diff
@@ -19,24 +19,34 @@ Opni currently features log anomaly detection for Kubernetes.
   * Ability to setup & send alerts (slack/email/etc) based on Opni log insights
 
 ----
-#### Try it out!
+### Install Opni in your Kubernetes cluster:
+* Download the `opnictl` binary from the [latest release](https://github.com/rancher/opni/releases/tag/v0.1.2)
```
kralicky, line 23: The url `https://github.com/rancher/opni/releases/latest` will redirect to the latest release.

README.md
```diff
@@ -19,24 +19,34 @@ Opni currently features log anomaly detection for Kubernetes.
...
+* Install Opni using `opnictl`
+  ```
+  opnictl install
+  opnictl create demo
+  ```
+  * Will look at `~/.kube/config` for your cluster's kubectl configuration file
+  * Cluster Hardware requirements: 4 vCPUs, 16GB RAM
+    * *2 Nvidia GPUs* required if you want the AI to learn from your workloads (recommended)
```
kralicky, line 31: Maybe we should mention vGPUs? You could use a single gpu split into 2 with VCS.

README.md
```diff
@@ -19,24 +19,34 @@ Opni currently features log anomaly detection for Kubernetes.
...
 ----
-#### Try it out!
+### Install Opni in your Kubernetes cluster:
+* Download the `opnictl` binary from the [latest release](https://github.com/rancher/opni/releases/tag/v0.1.2)
+* Install Opni using `opnictl`
+  ```
+  opnictl install
+  opnictl create demo
+  ```
+  * Will look at `~/.kube/config` for your cluster's kubectl configuration file
```
kralicky, line 29: This is not exactly correct - it uses the same logic that kubectl uses to determine which config and context to use. It would be more accurate to say "It will use your current kubeconfig context"

README.md
sanjay920: Yeah this is intended

README.md
sanjay920: We haven't tested in this setting so not comfortable stating vGPUs will work for our training/inferencing requirements.

kralicky (approved):

README.md
```diff
@@ -19,24 +19,34 @@ Opni currently features log anomaly detection for Kubernetes.
   * Ability to setup & send alerts (slack/email/etc) based on Opni log insights
 
 ----
```
wjimenez5271, line 21: What about putting a heading over this entire section titled "Getting Started with Opni" with a short explanation "here are two ways to start, 1. Add Opni to your existing cluster, or 2. Try Opni with our pre-built cluster and demo. Then follow with the detailed steps for each subheading using what you have?

sanjay920: addressed in https://github.com/rancher/opni/pull/86

PR #94, opened 2021-08-27
adding explicit NATS username/password env vars

sanjay920:


sanjay920:
> the build seems to be failing?

@dbason it looks like you updated the drone file?

dbason:
The drone update was just an adjustment to when we build the base images.

The build is currently failing on the e2e tests.  This is because they're currently pinned to the v0.1.2 versions of the microservices which don't include the new nats wrapper.  @sanjay920 @tybalex we can discuss a strategy for the new API in standup

PR #106, opened 2021-09-24
CLI updates

kralicky:


config/crd/bases/opni.io_logadapters.yaml
```diff
@@ -3001,13 +3001,13 @@ spec:
...
-                - RKE
-                - RKE2
-                - KubeAudit
+                - aks
+                - eks
+                - gke
+                - k3s
+                - rke
+                - rke2
+                - kubeaudit
```
dbason, line 3010: We can get rid of kubeaudit.  It isn't necessary for this.

config/crd/bases/opni.io_pretrainedmodels.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.5.0
+    controller-gen.kubebuilder.io/version: (devel)
```
dbason, line 7: This needs to be back on v0.5.0

controllers/logadapter_controller.go
```diff
@@ -70,21 +69,24 @@ func (r *LogAdapterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	// 			Controller:         pointer.Bool(true),
+	// 			BlockOwnerDeletion: pointer.Bool(true),
+	// 		},
+	// 	)
+	// 	logAdapter.Status.Phase = "Initializing"
+	// 	logAdapter.Status.Message = "Configuring Owner References"
+	// 	return ctrl.Result{
+	// 		Requeue: true,
+	// 	}, r.Update(ctx, &logAdapter)
+	// }
```
dbason, line 89: Let's just get rid of it rather than commenting it out

pkg/opnictl/commands/create_cluster.go
```diff
@@ -1,26 +1,381 @@
...
+							return fmt.Errorf("URL scheme must be http or https")
+						}
+						return nil
+					},
+				)),
+			)
+			if err != nil {
+				return nil, err
+			}
+		}
```
dbason, line 86: Should we just hardcode this seeing as we are responsible for producing and publishing the control plane model?

config/crd/bases/opni.io_logadapters.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.5.0
+    controller-gen.kubebuilder.io/version: (devel)
```
dbason, line 7: Set back to v0.5.0

config/crd/bases/opni.io_gpupolicyadapters.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.5.0
+    controller-gen.kubebuilder.io/version: (devel)
```
dbason, line 7: Set back to v0.5.0

config/crd/bases/demo.opni.io_opnidemoes.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.5.0
+    controller-gen.kubebuilder.io/version: (devel)
```
dbason, line 7: Set back to v0.5.0

dbason (changes requested):
It may also need a go mod tidy.  Once it's in I will need to update the e2e tests.

pkg/opnictl/commands/create_cluster.go
kralicky: This would be for airgap support

PR #110, opened 2021-09-28
Update ginkgo; replace build tags with ginkgo spec labels

kralicky:


test/e2e/cluster_test.go
```diff
@@ -1,6 +1,3 @@
-//go:build e2e
-// +build e2e
-
```
dbason, line 3: Should label the production api e2e tests so they get run

dbason (changes requested):

dbason (approved):

PR #149, opened 2021-11-08
Use forked elasticdump

dbason:


tybalex (approved):
This is to fix the es-dump error, right?

sanjay920 (approved):

dbason:
No longer required

PR #157, opened 2021-11-19
Metrics alerting

dbason:
This PR updates the metrics deployment to support scraping the predictions by prometheus.

The prometheusURL can now be inferred by checking externalURL if a reference to a prometheus is provided.
The operator will also deploy a ServiceMonitor and an alerting PrometheusRule if a prometheus reference is provided.


pkg/resources/opnicluster/services.go
```diff
@@ -600,19 +607,192 @@ func (r *Reconciler) gpuCtrlDeployment() (runtime.Object, reconciler.DesiredStat
...
 	return deployment, deploymentState(r.opniCluster.Spec.Services.Metrics.Enabled), nil
 }
 
+func (r *Reconciler) metricsService() (runtime.Object, reconciler.DesiredState, error) {
+	labels := r.serviceLabels(v1beta1.MetricsService)
+	service := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      labels[resources.AppNameLabel],
+			Namespace: r.opniCluster.Namespace,
+			OwnerReferences: []metav1.OwnerReference{
```
kralicky, line 651: Instead of setting this manually, we should use `ctrl.SetControllerReference()` which also sets the Controller field in the owner reference (also applies to other usages in this file)

pkg/resources/opnicluster/services.go
```diff
@@ -600,19 +607,192 @@ func (r *Reconciler) gpuCtrlDeployment() (runtime.Object, reconciler.DesiredStat
...
+		err := r.client.Get(r.ctx, types.NamespacedName{
+			Name:      r.opniCluster.Spec.Services.Metrics.PrometheusReference.Name,
+			Namespace: r.opniCluster.Spec.Services.Metrics.PrometheusReference.Namespace,
+		}, prometheus)
+		if err != nil {
+			return prometheusRule, deploymentState(r.opniCluster.Spec.Services.Metrics.Enabled), err
+		}
+		if prometheus.Spec.RuleNamespaceSelector == nil {
+			prometheusRule.SetNamespace(prometheus.Namespace)
+			err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
kralicky, line 755: Missing check for this error

pkg/resources/opnicluster/services.go
```diff
@@ -600,19 +607,192 @@ func (r *Reconciler) gpuCtrlDeployment() (runtime.Object, reconciler.DesiredStat
...
+					return err
+				}
+				if r.opniCluster.Status.PrometheusRuleNamespace != prometheus.Namespace {
+					r.opniCluster.Status.PrometheusRuleNamespace = prometheus.Namespace
+					err = r.client.Status().Update(r.ctx, r.opniCluster)
+					if err != nil {
+						return err
+					}
+				}
+				if r.opniCluster.DeletionTimestamp == nil && !controllerutil.ContainsFinalizer(r.opniCluster, prometheusRuleFinalizer) {
```
kralicky, line 766: AddFinalizer does the contains check for you

pkg/resources/opnicluster/services.go
```diff
@@ -600,19 +607,192 @@ func (r *Reconciler) gpuCtrlDeployment() (runtime.Object, reconciler.DesiredStat
...
+						},
+					},
+				},
+			},
+		},
+	}
+	// Fetch Prometheus resource to calculate namespace and match labels for rules
+	if r.opniCluster.Spec.Services.Metrics.PrometheusReference != nil {
+		prometheus := &monitoringv1.Prometheus{}
+		err := r.client.Get(r.ctx, types.NamespacedName{
```
kralicky, line 746: you can use `client.ObjectKeyFromObject()` instead of manually filling out the NamespacedName

controllers/suite_test.go
```diff
@@ -263,3 +277,10 @@ func buildCluster(opts opniClusterOpts) *v1beta1.OpniCluster {
 		},
 	}
 }
+
+func generateSHAID(name string, namespace string) string {
+	hash := sha1.New()
+	hash.Write([]byte(name + namespace))
+	sum := hash.Sum(nil)
+	return fmt.Sprintf("%x", sum[:3])
```
kralicky, line 285: `hex.EncodeToString` probably better here (although it doesnt really make a difference)

kralicky (changes requested):

pkg/resources/opnicluster/services.go
dbason: We don't use this for any of the other resources so I've stuck to what we are already using.  If we want to change this we should do it for all to keep them consistent

pkg/resources/opnicluster/services.go
dbason: This doesn't implement client.Object so we can't use that.

pkg/resources/opnicluster/services.go
kralicky: We may be setting them incorrectly in some places - SetControllerReference sets `Controller: true` in the owner reference. If that field is not set I believe it will prevent things like finalizers from running and it also affects deletion propagation.

kralicky (approved):

PR #160, opened 2021-11-24
Metric service  ISM policy

tybalex:
this PR defines the ISM policy for opni metric service.
https://zube.io/rancher/opni/c/276


pkg/resources/opnicluster/elastic/indices/opni_settings.go
```diff
@@ -17,6 +17,10 @@ const (
 	drainStatusIndexPrefix       = "opni-drain-model-status-v0.1.3"
 	drainStatusIndexAlias        = "opni-drain-model-status"
 	drainStatusIndexTemplateName = "opni-drain-model-status_rollover_mapping"
+	metricPolicyName             = "opni-metric-policy"
+	metricIndexPrefix            = "opni-metric-v0.1.3"
```
dbason, line 21: We should probably set this to `opni-metric-v0.3.0` to match the release

dbason (changes requested):
Looks good, just a minor nit

pkg/resources/opnicluster/elastic/indices/opni_settings.go
tybalex: add a const releaseVersion=v0.3.0 to match our next release.

dbason (changes requested):
I wouldn't change the existing index names, because then we have to do more manual testing to confirm that an existing cluster gets upgraded correctly.  It should but we do have to verify it.  Also if you change the index names we have to modify the existing e2e tests to verify them - https://github.com/rancher/opni/blob/main/test/e2e/cluster_test.go#L314-L319

It may also be worthwhile adding the new index and template to the e2e tests.

tybalex:
> I wouldn't change the existing index names, because then we have to do more manual testing to confirm that an existing cluster gets upgraded correctly. It should but we do have to verify it. Also if you change the index names we have to modify the existing e2e tests to verify them - https://github.com/rancher/opni/blob/main/test/e2e/cluster_test.go#L314-L319
> 
> It may also be worthwhile adding the new index and template to the e2e tests.

This makes sense to me. Let me revert it back then. Maybe we can update these index names after Dec 10 release? that way we can have time to test it.

dbason (approved):

PR #209, opened 2022-05-09
Run logging controllers in agent

dbason:


kralicky (approved):

pkg/opni/common/options.go
```diff
@@ -58,3 +60,25 @@ func LoadDefaultClientConfig() {
 	APIConfig, RestConfig, K8sClient = cliutil.CreateClientOrDie(
 		append(MaybeContextOverride(), cliutil.WithExplicitPath(ExplicitPathFlagValue))...)
 }
+
+func GetClientOrDie(inCluster bool) client.Client {
```
kralicky, line 64: We should merge this with code in util/cluster.go

pkg/opni/root.go
```diff
@@ -30,5 +33,8 @@ func BuildRootCmd() *cobra.Command {
 }
 
 func Execute() {
-	BuildRootCmd().Execute()
+	if err := BuildRootCmd().Execute(); err != nil {
```
kralicky, line 36: Check to make sure this doesn't cause errors to be printed twice

PR #228, opened 2022-05-25
Add Opensearch ingest plugin and configurable index retention

dbason:
Also adds OpenSearchCluster as an owned resource to some controllers to configure reconciliation if that changes.


tybalex (approved):
look good to me

PR #233, opened 2022-05-27
dependency updates

kralicky:


images/python/requirements.txt
```diff
@@ -1,12 +1,12 @@
...
 boto3==1.17.45
 botocore==1.20.45
 cachetools==4.2.1
 elasticsearch[async]==7.13.0
-fastapi==0.63.0
+fastapi==0.65.2
 jsonpickle==1.5.1
 kubernetes==12.0.1
-numpy==1.20.3
+numpy==1.21
```
sanjay920, line 9: let's not update these unless we need to. they can result in unexpected changes/breaks. everything else looks fine from my end

sanjay920 (changes requested):

PR #234, opened 2022-05-30
Update mappings and add index upgrade logic

dbason:


sanjay920 (approved):

pkg/resources/opnicluster/elastic/indices/opni_settings.go
```diff
@@ -24,11 +24,14 @@ const (
...
 	normalIntervalIndexName      = "opni-normal-intervals"
 	kibanaDashboardVersionDocID  = "latest"
-	kibanaDashboardVersion       = "v0.4.0"
+	kibanaDashboardVersion       = "v0.5.0"
 	kibanaDashboardVersionIndex  = "opni-dashboard-version"
 )
 
 var (
+	OldIndexPrefixes = []string{
+		"logs-v0.1.3*",
```
tybalex, line 33: is v0.1.3 the only version that needs upgrade? or this logic should probably do upgrade for any version that older than current version.

dbason: It's the only version that will exist

PR #241, opened 2022-06-10
Introduction of service to update logs in Opensearch.

AmartC:
This PR introduces the Opensearch Updating Service which will now receive logs through Nats from the DRAIN and Inferencing services and update to Opensearch from there.

### DIFF
--- a/aiops/opni-opensearch-update-service/Readme.md
+++ b/aiops/opni-opensearch-update-service/Readme.md
@@ -1,0 +1,5 @@
+## Testing the Opensearch Updating Service
+Simply run 
+```
+python3 test_opensearch_update.py
+```
--- a/aiops/opni-opensearch-update-service/Readme.md
+++ b/aiops/opni-opensearch-update-service/Readme.md
@@ -1,5 +1,0 @@
-## Testing the Opensearch Updating Service
-Simply run 
-```
-python3 test_opensearch_update.py
-```
--- a/aiops/opni-opensearch-update-service/opensearch-update-service/app/Dockerfile
+++ b/aiops/opni-opensearch-update-service/opensearch-update-service/app/Dockerfile
@@ -1,0 +1,4 @@
+FROM rancher/opni-python-base:3.8
+COPY ./main.py .
+
+CMD ["python", "./main.py"]
--- a/aiops/opni-opensearch-update-service/opensearch-update-service/app/Dockerfile
+++ b/aiops/opni-opensearch-update-service/opensearch-update-service/app/Dockerfile
@@ -1,4 +1,0 @@
-FROM rancher/opni-python-base:3.8
-COPY ./main.py .
-
-CMD ["python", "./main.py"]
--- a/aiops/opni-opensearch-update-service/opensearch-update-service/main.py
+++ b/aiops/opni-opensearch-update-service/opensearch-update-service/main.py
@@ -1,0 +1,148 @@
+# Standard Library
+import asyncio
+import json
+import logging
+import os
+
+# Third Party
+import pandas as pd
+from elasticsearch import AsyncElasticsearch, TransportError
+from elasticsearch.exceptions import ConnectionTimeout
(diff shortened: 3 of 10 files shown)


aiops/opni-opensearch-update-service/Readme.md
```diff
@@ -0,0 +1,5 @@
+## Testing the Opensearch Updating Service
+Simply run 
+```
+python3 test_opensearch_update.py
```
kralicky, line 4: Tests should be run using `pytest` - Also please rename to README.md if you want to keep this file

kralicky (changes requested):
Please add the following files in the `aiops` folder:
- `.gitignore` with python rules
- `requirements.txt` 
- `.coveragerc` with the following contents:
```
[run]
omit = 
  */test_*
```

aiops/opni-opensearch-update-service/opensearch-update-service/main.py
```diff
@@ -0,0 +1,148 @@
...
+
+async def receive_logs(queue):
+    es = await setup_es_connection()
+    while True:
+        df = await queue.get()
+        await update_logs(es, df)
+
+async def update_logs(es, df):
+    # This function will be updating Opensearch logs which were inferred on by the DRAIN model.
+    model_keywords_dict = {"drain": {"Normal":  ["_id", "_op_type", "_index", "masked_log", "drain_pretrained_template_matched", "inference_model"],
```
tybalex, line 90: we should update anomaly level for Normal logs too.

aiops/opni-opensearch-update-service/opensearch-update-service/test_opensearch_update.py
```diff
@@ -0,0 +1,42 @@
+import asyncio
+import pandas as pd
+import unittest
+from unittest import IsolatedAsyncioTestCase
+from main import doc_generator
+
+class TestStringMethods(IsolatedAsyncioTestCase):
```
tybalex, line 9: let's put the test scripts under directory `test`

ci.cue
```diff
@@ -355,5 +360,24 @@ dagger.#Plan & {
 				}
 			}
 		}
+		aiops: {
```
tybalex, line 363: is it the right way to do? looks like it builds everything under `aiops` but don't we only want the open search-update-service

aiops/opni-opensearch-update-service/opensearch-update-service/main.py
AmartC: the default anomaly level is set to Normal though by the ingest plugin so we don't need to update that again right?

tybalex: https://github.com/tybalex/opni-ingest-plugin/pull/9
let's change the default value to ""

aiops/requirements.txt
```diff
@@ -0,0 +1,3 @@
+elasticsearch[async]==7.13.0
```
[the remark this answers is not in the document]
kralicky: this was to have a minimal set of requirements for code under the new aiops folder. e.g. from a blank environment, what packages to install just to test this code. it will be better to have this here as more code is added.

ci.cue
kralicky: This was just sort of an example for the new code. It is not replacing the current system yet.

aiops/opni-opensearch-update-service/opensearch-update-service/app/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM rancher/opni-python-base:3.8
```
tybalex, line 1: Do we need Dockerfiles and if yes why it is not in the root dir of this service?

kralicky: No, we should not check in any new dockerfiles.

tybalex (approved):
lg

kralicky (approved):

PR #294, opened 2022-06-22
end to end test setup + extras

kralicky:
- e2e test infra using pulumi
- e2e test running from dagger
- one basic e2e test as a starting point
- several bugfixes
- switch http router library from fiber to gin
- add opentelemetry trace instrumentation for http/grpc, still needs more manual span setup
- some logging fixes for plugins


controllers/gateway_controller_test.go
```diff
@@ -10,15 +10,14 @@ import (
...
-	"github.com/rancher/opni/pkg/test"
 	"github.com/rancher/opni/pkg/util"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var _ = Describe("Gateway Controller", Ordered, Label(test.Controller, test.Slow), func() {
+var _ = Describe("Gateway Controller", Ordered, Label("controller", "slow"), func() {
```
dbason, line 20: How come we're moving away from constants?

dbason:
@kralicky the testing has picked up a data race which you will need to review

controllers/gateway_controller_test.go
kralicky: The `ginkgo labels` command is used to filter tests by label, but it can only identify string constants since it works by parsing the AST. Running ginkgo with `--label-filter` (like we have been doing until now) won't work because it has to compile the tests to check the filter, and will delete the test binaries once it determines they should be skipped. This means we can't run two disjoint sets of tests at the same time, because one will end up deleting a binary the other set needs before it can run.

dbason (dismissed):

dbason (approved):

PR #322, opened 2022-06-30
Fix sloth 

alexandreLamarre:
Fixing sloth in the best way possible : removing sloth (kind of)

Once I started trying to generate alerts with sloth + my custom OpenSLO v1 format IR, I realized using the sloth generators weren't going to work at all for our use-case.

1) they are tied to OO where the methods are predicated on reading from the filesystem
2) Confusing function factories and factory configs which were tied to very specific yaml input formats

In essence, I stripped the sloth generator down to the basics and repurposed the raw core functions into the plugin package `plugins/slo/pkg/slo` so we can have more control over how to generate alerts & recording rules

### DIFF
--- a/plugins/slo/pkg/slo/doc.go
+++ b/plugins/slo/pkg/slo/doc.go
@@ -1,0 +1,16 @@
+/*
+   Copyright [2021] [Xabier Larrakoetxea]
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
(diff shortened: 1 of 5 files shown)


plugins/slo/pkg/slo/doc.go
```diff
@@ -0,0 +1,16 @@
+/*
```
kralicky, line 1: I would add a comment like this here above the license:

```
This package contains code redistributed from Sloth (github url) under the following license:
```

kralicky (approved):

PR #323, opened 2022-06-30
Slo metric discovery (regex in promQL)

alexandreLamarre:
Integrate the regex into promql queries when fetching potential downstream matches to the preconfigured metrics.

Returns codes.NotFound when no metric matches the preconfigured metric.


plugins/slo/pkg/slo/metrics.go
```diff
@@ -134,7 +143,74 @@ func (p *PrometheusQueryImpl) IsHistogram() bool {
...
+	res := make([]string, 0)
+	data := resp.GetData()
+	lg.Debug(fmt.Sprintf("Received service data:\n %s from cluster %s ", string(data), metricRequest.ClusterId))
+	q, err := unmarshal.UnmarshallPrometheusResponse(data)
+	if err != nil {
+		return "", err
+	}
+	switch q.V.Type() {
+	case model.ValVector:
+		var vv model.Vector = q.V.(model.Vector)
```
kralicky, line 189: Simplify as `vv := q.V.(model.Vector)`

plugins/slo/pkg/slo/metrics.go
```diff
@@ -134,7 +143,74 @@ func (p *PrometheusQueryImpl) IsHistogram() bool {
...
+		Tenants: []string{metricRequest.ClusterId},
+		Query:   strings.TrimSpace(query.String()),
+	})
+	if err != nil {
+		lg.Error(fmt.Sprintf("Failed to query cluster %v: %v", metricRequest.ClusterId, err))
+		return "", err
+	}
+	res := make([]string, 0)
+	data := resp.GetData()
+	lg.Debug(fmt.Sprintf("Received service data:\n %s from cluster %s ", string(data), metricRequest.ClusterId))
```
kralicky, line 182: Should keep logs to a single line

plugins/slo/pkg/apis/slo/slo.proto
```diff
@@ -53,7 +53,7 @@ service SLO {
   }
 
   // Can this metric run with this service & cluster? No == error
-  rpc GetMetric(MetricRequest) returns (google.protobuf.Empty) {
+  rpc AssignMetric(MetricRequest) returns (Metric) {
```
kralicky, line 56: the name `AssignMetric` suggests this method will have side effects, but it looks like this is more of a query.

plugins/slo/pkg/slo/metrics.go
```diff
@@ -39,21 +41,27 @@ type goodQueryPlaceholder struct {
 	Filter   string
 }
 
+type metricTemplate struct {
+	NameRegex string
+	ServiceId string
+}
+
 func InitMetricList() {
```
kralicky, line 49: Would this be better as an `init()` method?

plugins/slo/pkg/slo/metrics.go
```diff
@@ -134,7 +143,74 @@ func (p *PrometheusQueryImpl) IsHistogram() bool {
 	return p.isHistogram
 }
 
-// @Note: Assumption is that JobID is valid
+func selectBestMatch(metrics []string) string {
+	min := int(^uint(0) >> 1) // largest int
```
kralicky, line 147: You can use `math.MaxInt` for this value

plugins/slo/pkg/slo/metrics.go
```diff
@@ -134,7 +143,74 @@ func (p *PrometheusQueryImpl) IsHistogram() bool {
...
+	case model.ValVector:
+		var vv model.Vector = q.V.(model.Vector)
+		if len(vv) == 0 {
+			err := status.Error(codes.NotFound,
+				fmt.Sprintf("No assignable metric '%s' for service '%s' in cluster '%s' ",
+					metricRequest.Name, metricRequest.ServiceId, metricRequest.ClusterId))
+			return "", err
+		}
+		for _, v := range vv {
+			res = append(res, (string)(v.Metric["__name__"]))
```
kralicky, line 197: Prefer `type(value)` over `(type)(value)` where possible

test/plugins/slo/slo_test.go
```diff
@@ -4,12 +4,15 @@ import (
 	"context"
 	"time"
 
+	"github.com/gogo/status"
```
kralicky, line 7: Keep an eye out for this, since the IDE will often autocomplete this package (or grpc.go4.org) instead of `google.golang.org/grpc/status`, which is the correct one for this project. Make sure to run `go mod tidy` after and check that it is no longer being imported.

kralicky (changes requested):

plugins/slo/pkg/slo/metrics.go
alexandreLamarre: I think putting this as part of the `plugin` startup code would be best. Is there an existing place to do that

alexandreLamarre: Oh I see `init` has a special meaning in go, yeah let me put the contents of InitMetricList into `init`, but also I'll put the metric list in the (k,v) store when we first set the storage in the SLO plugin

plugins/slo/pkg/slo/metrics.go
alexandreLamarre: Done!

plugins/slo/pkg/slo/metrics.go
alexandreLamarre: Done!

plugins/slo/pkg/apis/slo/slo.proto
alexandreLamarre: Renamed to `GetMetricId`, since we are fetching the exact id of a metric using preconfigured metric name, e.g. `uptime` will try to match the closest metric to prometheus' `up` on a service, etc...

plugins/slo/pkg/slo/metrics.go
alexandreLamarre: Did a minor refactor, put `InitMetricList` into the special `init` function in this module, and removed references to it in the api package, preferring to set it once (and not check in each method if it is initialized) when we set the storage future for the SLO plugin

plugins/slo/pkg/slo/metrics.go
alexandreLamarre: Done!

plugins/slo/pkg/slo/metrics.go
alexandreLamarre: Done!

test/plugins/slo/slo_test.go
alexandreLamarre: that's a weird one, but it is fixed now

kralicky (approved):

PR #361, opened 2022-07-06
Slo CRUD Implementation

alexandreLamarre:
- Simplify the data an SLO holds into : 
```
  string name = 1;
  string datasource = 2;
  string monitorWindow = 3; // 7d, 28d or 30d
  string budgetingInterval = 4; // in the format 5m, 1h etc..
  Target target = 5;
  repeated Label labels = 6;
  repeated Alert alerts = 7;
```
- Implement CRUD SLO api for monitoring datasources 

------

- Fixed bugs in yesterday's implementation:
  - Update works in multi cluster env, (was leaving old rule group in original cluster)
  - CreateSLO / UpdateSLO was not applying metadata & alerting rule groups, just recording rule groups 
  - DeleteSLO actually deletes the sub group rules, not a generic non-existent rule group with the SLO id


alexandreLamarre:
Note: Uptime pre-configured metric still does not apply the ranges 5m, 10m, 30m properly, requires an overhaul of recording rule generation

pkg/slo/query/query_test.go
```diff
@@ -17,7 +17,7 @@ var _ = Describe("Converting ServiceLevelObjective Messages to Prometheus Rules"
 
 	When("We define pre-configured metrics in def.go", func() {
 		It("Should produce valid templates and regex at compile time", func() {
-			Expect(query.AvailableQueries).To(HaveLen(3))
+			Expect(query.AvailableQueries).ToNot(HaveLen(0)) // > 0 available pre configured metrics
```
kralicky, line 20: nit: use `NotTo` instead of `ToNot` (autocomplete always suggests ToNot for some reason)

plugins/slo/pkg/slo/validate.go
```diff
@@ -16,18 +21,33 @@ func matchEnum(target string, enum []string, returnErr error) error {
...
 	}
-	if err := validateAlert(slo.GetAlerts()); err != nil {
-		return err
+	// Expected monitor window
+	validWindows := []string{"7d", "28d", "30d"}
+	if !slices.Contains(validWindows, slorequest.SLO.GetMonitorWindow()) {
+		return shared.ErrInvalidMonitorWindow
+	}
+
+	budgetingIntervalRegex := regexp.MustCompile(`(^\d+)[m]$`)
```
kralicky, line 40: If you edit the proto field type (see comment in slo.proto) you won't need this parsing logic anymore. you can also allow arbitrary durations for monitoring windows, and/or compare them against strongly typed duration values instead of strings.

plugins/slo/pkg/slo/interfaces.go
```diff
@@ -0,0 +1,64 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+var datasourceToImpl map[string]SLOImpl = make(map[string]SLOImpl)
+
+func init() {
+	datasourceToImpl[shared.MonitoringDatasource] = SLOMonitoring{}
+}
+
+type SLOImpl interface {
```
kralicky, line 20: the naming here is a bit weird. I'd avoid the name `impl` for interfaces or interface methods. A better name for this interface might be `Backend` or `Store` (such that the full name is `slo.Backend` or `slo.Store` which reads better in the context of the package

plugins/slo/pkg/slo/interfaces.go
```diff
@@ -0,0 +1,64 @@
...
+	New(req proto.Message, p *Plugin, ctx context.Context, lg hclog.Logger) SLOImpl
```
kralicky, line 21: This method should not be inside the interface, rather you should have two constructors e.g. `NewMonitoringStore` and `NewLoggingStore` which each return some type that implements the interface

plugins/slo/pkg/slo/interfaces.go
```diff
@@ -0,0 +1,64 @@
...
+	// This method signature has to handle storage of the SLO in the KVStore
+	// Since there can be partial successes
+	createSLOImpl([]v1.SLO) (*apis.CreatedSLOs, error)
```
kralicky, line 24: these methods should probably be exported, and I would suggest renaming them like `Create` `Update` `Delete` etc.

plugins/slo/pkg/slo/openslo.go
```diff
@@ -99,17 +101,14 @@ func ParseToIndicator(slo *api.ServiceLevelObjective, service *api.Service, ctx
...
-			DisplayName:     slo.GetName() + "-target" + strconv.Itoa(i),
-			Target:          float64(target.GetValueX100()) / 100,
-			TimeSliceWindow: slo.GetBudgetingInterval(),
-		}
-		objectives = append(objectives, newObjective)
+func ParseToObjective(slo *api.ServiceLevelObjective, ctx context.Context, lg hclog.Logger) oslov1.Objective {
+	target := slo.GetTarget()
+	newObjective := oslov1.Objective{
+		DisplayName:     slo.GetName() + "-target",
+		Target:          float64(target.GetValueX100()) / 100,
```
kralicky, line 108: I'd recommend adding this math into a helper function added to the Target type (inside the apis/slo package). like GetValueNormalized or something

plugins/slo/pkg/slo/store.go
```diff
@@ -0,0 +1,202 @@
...
+	return &CortexRuleWrapper{
+		recording: string(rrecording),
+		metadata:  string(rmetadata),
+		alerts:    string(ralerts),
+	}, nil
+}
+
+// Cortex applies rule groups individually
+func applyCortexSLORules(p *Plugin, cortexRules *CortexRuleWrapper, service *sloapi.Service, existingId string, ctx context.Context, lg hclog.Logger) error {
+	var anyError error
```
kralicky, line 99: you should probably do this in a loop, it is hard at first glance to tell what's going on here

plugins/slo/pkg/apis/slo/slo.proto
```diff
@@ -121,65 +107,36 @@ message Alert {
...
-// TODO : rework this
-message SLOStatus {
-  core.Reference slo = 1;
-  int64 timestamp = 2;
-  Evaluation status = 3;
-  Evaluation budget = 4;
+  string name = 1;
+  string datasource = 2;
+  string monitorWindow = 3; // 7d, 28d or 30d
+  string budgetingInterval = 4; // in the format 5m, 1h etc..
```
kralicky, line 113: Use google.protobuf.Duration for the these fields, so that it will turn the field into native time structs

plugins/slo/pkg/apis/slo/slo.proto
```diff
@@ -121,65 +107,36 @@ message Alert {
...
-  core.Reference slo = 1;
-  State state = 2;
+message Services {
+  repeated Service items = 1;
 }
 
-// Contains all the SLO data like, error budget, burn rate, etc...
-message State {
-  bytes data = 1;
+message CreatedSLOs {
```
kralicky, line 128: Use `core.ReferenceList` instead of a custom type

plugins/slo/pkg/apis/slo/slo.proto
```diff
@@ -121,65 +107,36 @@ message Alert {
...
 
-message FormulaList {
-  repeated Formula items = 1;
-}
-
-message Evaluation {
-  double rate = 1;
-  double period = 2;
+message Target {
+  uint64 ValueX100 = 1;
```
kralicky, line 139: rename with lowercase v (for consistent json key naming)

plugins/slo/pkg/apis/slo/slo.proto
```diff
@@ -196,4 +153,24 @@ message MetricRequest {
...
+message MetricList {
+  repeated Metric items = 1;
+}
+
+message ServiceLevelObjectiveList {
+  repeated SLOImplData items = 1;
+}
+
+enum SLOStatusState {
+  SLO_STATUS_NO_DATA = 0;
```
kralicky, line 167: These enum names should be CamelCase, protobuf adds extra prefixes onto the constants it generates so the names won't clash

plugins/slo/pkg/slo/api.go
```diff
@@ -39,50 +36,33 @@ func list[T proto.Message](kvc system.KVStoreClient[T], prefix string) ([]T, err
...
+	if _, ok := datasourceToImpl[datasource]; !ok {
+		return shared.ErrInvalidDatasource
 	}
+	return nil
+}
 
-	if err := ValidateInput(slo); err != nil {
+func (p *Plugin) CreateSLO(ctx context.Context, slorequest *sloapi.CreateSLORequest) (*sloapi.CreatedSLOs, error) {
+	lg := p.logger
+	lg.Debug(fmt.Sprintf("%d", len(datasourceToImpl)))
```
kralicky, line 48: these logs are probably not needed

plugins/slo/pkg/slo/api.go
```diff
@@ -39,50 +36,33 @@ func list[T proto.Message](kvc system.KVStoreClient[T], prefix string) ([]T, err
...
-	}
-
-	// Put in k,v store only if everything else succeeds
-	if err := p.storage.Get().SLOs.Put(path.Join("/slos", slo.Id), slo); err != nil {
+	osloSpecs, err := ParseToOpenSLO(slorequest, ctx, p.logger)
+	if err != nil {
 		return nil, err
 	}
-	return &emptypb.Empty{}, nil
+	lg.Debug(fmt.Sprintf("Length of osloSpecs : %d", len(osloSpecs)))
```
kralicky, line 60: this message should be more descriptive

test/plugins/slo/slo_test.go
```diff
@@ -2,23 +2,93 @@ package plugins_test
...
+	}
+
+	if err := expectRuleGroupToNotExist(adminClient, ctx, tenant, groupName+slo.AlertRuleSuffix); err != nil {
+		anyError = err
+	}
+	return anyError
+}
+
+func expectRuleGroupToExist(adminClient cortexadmin.CortexAdminClient, ctx context.Context, tenant string, groupName string) error {
+	for i := 0; i < 10; i++ {
```
kralicky, line 56: these helper functions are good, but you should implement them using `Eventually()`.

kralicky (changes requested):

plugins/slo/pkg/slo/interfaces.go
alexandreLamarre: You're right, I think here my idea was because I was using the message `SLOImplData` as the data stored in the backend. Something like : "This is all the data that an SLO implementation needs". 

I think I should also rename the message to match the new interface name

plugins/slo/pkg/slo/api.go
alexandreLamarre: Done

test/plugins/slo/slo_test.go
alexandreLamarre: Refactored them using `sync.WaitGroup` and `go func(){}` instead of `Eventually()` since I had trouble refactoring them correctly to an Eventually matcher

plugins/slo/pkg/slo/api.go
alexandreLamarre: Removed

plugins/slo/pkg/slo/interfaces.go
alexandreLamarre: Done

plugins/slo/pkg/slo/interfaces.go
alexandreLamarre: Removed,  and added `NewMonitoringStore`,

but added an extra interface method called `WithCurrentRequest` which sets the information needed to process the current request. 

Before resolving, I'm going to implement the entire logging interface to return grpc UNIMPLEMENTED error codes.

plugins/slo/pkg/slo/interfaces.go
alexandreLamarre: Renamed the interfaces based on the suggested naming conventions, still need to rename the `SloImplData` Message

pkg/slo/query/query_test.go
alexandreLamarre: Done

plugins/slo/pkg/slo/interfaces.go
alexandreLamarre: Done

plugins/slo/pkg/slo/validate.go
alexandreLamarre: Converted budgeting interval to google's protobuf.Duration

plugins/slo/pkg/apis/slo/slo.proto
alexandreLamarre: Done

plugins/slo/pkg/apis/slo/slo.proto
alexandreLamarre: Done

plugins/slo/pkg/apis/slo/slo.proto
alexandreLamarre: Done, except the monitorWindow which is intended to be an enum.

plugins/slo/pkg/slo/interfaces.go
alexandreLamarre: Done, Actually decided to remove `NewLoggingStore`, since there is not even code that fetches the logging plugin anyways

plugins/slo/pkg/slo/openslo.go
alexandreLamarre: Done

plugins/slo/pkg/slo/store.go
alexandreLamarre: Done

plugins/slo/pkg/apis/slo/slo.proto
alexandreLamarre: Done

kralicky (approved):

PR #366, opened 2022-07-11
Refactor Service Discovery to agnostic Service Discovery backend interfaces

alexandreLamarre:
Similar design strategy to how we implement `SLOStore` that handles abstraction over datasource specific SLO CRUD, we now have `ServiceBackend` that handles abstraction over datasource specific service discovery.


test/plugins/slo/slo_test.go
```diff
@@ -133,7 +133,7 @@ var _ = Describe("Converting ServiceLevelObjective Messages to Prometheus Rules"
 
 	When("The SLO plugin starts", func() {
 		It("should be able to discover services from downstream", func() {
-
+			time.Sleep(time.Second * 10) // Wait for cortex to be registered
```
kralicky, line 136: probably need a better way to do this, 10 second sleep is kinda long. can you poll an api to wait instead?

kralicky (changes requested):

test/plugins/slo/slo_test.go
alexandreLamarre: In my previous PR this wait wasn't needed, but (without any of the features in the current branch) it still required 10 second wait, otherwise the tests would fail because the admin plugin wasn't set yet. Any idea what could be causing this?

PR #370, opened 2022-07-12
Cluster capability uninstall

### LINKED ISSUE #266: Allow purging metric data when deleting a cluster
Currently there is no way to purge metric data from long-term storage for a cluster. This is a necessary feature, and supported by cortex.

kralicky:
Closes #266


dbason (commented):
Can't see anything obviously wrong.  Just need to make sure all tests are green, and we should wait until we've cut the release.

sanjay920 (approved):

PR #397, opened 2022-07-20
Create separate task for pypi build that is not included in Drone

AmartC:
This PR creates a separate task for pushing to the Pypi repository: opni-proto.


dbason (changes requested):
You need to update drone to only do the opensearchUpdateService action

ci.cue
```diff
@@ -57,7 +57,6 @@ dagger.#Plan & {
 			"bin": write: contents:             actions.build.bin
 			"web/dist": write: contents:        actions.web.dist
 			"cover.out": write: contents:       actions.test.export.files["/src/cover.out"]
-			"aiops/apis/dist": write: contents: actions.aiops.sdist.output
```
kralicky, line 60: I'd recommend keeping this here, it could be useful

kralicky (changes requested):

dbason (approved):

kralicky (approved):

PR #398, opened 2022-07-20
Fix upgrade responder and add extra info

### LINKED ISSUE #392: Review upgrade-responder telemetry and add it to the agent
Check that it is working with new builds as numbers have fallen off significantly.

Upgrade responder also needs to be able to track agent installs, and stats should be separated across agent and cluster installs.

dbason:
PR make the following changes:
 - Removed the dirty repo checks for versioning because they fail when running in the dagger environment.
 - Moved disable usage to a global persistent flag.
 - Added upgrade responder to the agent, and added extra info to differentiate the two.

Fixes #392

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -182,12 +182,6 @@ func init() {
 	}
 
 	// get version info
-	dirty := true
-	if output, err := sh.Output("git", "status", "--porcelain", "--untracked-files=no"); err != nil {
-		panic(err)
-	} else if strings.TrimSpace(output) == "" {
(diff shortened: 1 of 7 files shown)


dbason:
Primarily upgrade-responder is for telemetry (as a side node we need to keep the version in the server up to date).  I don't think this will impact that in any way and this is an explicit ask from @sanjay920

kralicky:
lgtm

kralicky (approved):

PR #401, opened 2022-07-20
Update publish version tag for opensearch updating service image

AmartC:
This PR updates the publish version tag for the opni-opensearch-update-service image for RC 6.


dbason (changes requested):
As discussed this is the wrong thing to change.  Need to add the `TAG` environment variable to the step in drone

dbason (approved):

PR #403, opened 2022-07-20
Check for nil pointer in gw.Spec.Alerting

alexandreLamarre:
in `resources/gateway/configmap.go`, need to check for r.gw.Spec.Alerting is nil, since we decided to specify that part of the spec as optional (and therefore as a pointer)

### DIFF
--- a/pkg/resources/gateway/configmap.go
+++ b/pkg/resources/gateway/configmap.go
@@ -44,14 +44,19 @@ func (r *Reconciler) configMap() (resources.Resource, error) {
 			Storage: cfgv1beta1.StorageSpec{
 				Type: r.gw.Spec.StorageType,
 			},
-			Alerting: cfgv1beta1.AlertingSpec{
-				Endpoints: []string{
-					fmt.Sprintf("opni-alerting:%d", r.gw.Spec.Alerting.Port),
-				},
-			},
 		},
 	}
 	gatewayConf.Spec.SetDefaults()
+
+	if r.gw.Spec.Alerting == nil {
+		gatewayConf.Spec.Alerting = cfgv1beta1.AlertingSpec{
+			Endpoints: []string{"opni-alerting:9093"},
+		}
+	} else {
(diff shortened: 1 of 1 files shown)


sanjay920 (dismissed):

pkg/resources/gateway/configmap.go
```diff
@@ -44,15 +44,20 @@ func (r *Reconciler) configMap() (resources.Resource, error) {
...
+	if r.gw.Spec.Alerting == nil {
+		gatewayConf.Spec.Alerting = cfgv1beta1.AlertingSpec{
+			Endpoints: []string{"opni-alerting:9093"},
+		}
+	} else {
+		gatewayConf.Spec.Alerting = cfgv1beta1.AlertingSpec{
+			Endpoints: []string{fmt.Sprintf("opni-alerting:%d", r.gw.Spec.Alerting.Port)},
+		}
+	}
+
```
dbason, line 60: You can put it into a function e.g.
```go
Alerting: func() cfgv1beta1.AlertingSpec {
    if r.gw.Spec.Alerting == nil {
		return gatewayConf.Spec.Alerting = cfgv1beta1.AlertingSpec{
			Endpoints: []string{"opni-alerting:9093"},
		}
	return gatewayConf.Spec.Alerting = cfgv1beta1.AlertingSpec{
			Endpoints: []string{fmt.Sprintf("opni-alerting:%d", r.gw.Spec.Alerting.Port)},
		}
}()
```

dbason: But with proper spacing

dbason (changes requested):

pkg/resources/gateway/configmap.go
alexandreLamarre: On it

alexandreLamarre: Done

dbason (approved):

Issue #317, opened 2022-06-29
Allow user to install opni into any namespace


sanjay920:
Right now user is limited to `opni-cluster-system`

PR #410, opened 2022-07-21
Fix reconciler error with correct ns name

alexandreLamarre:
couple of `OPNI_SYSTEM_NAMESPACE` env var were still in use, when  we made the change for opni-manager to use `POD_NAMESPACE` they were leftover

See https://github.com/rancher/opni/pull/327 & https://github.com/rancher/opni/issues/317

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -32,7 +32,7 @@ spec:
           image: controller
           name: manager
           env:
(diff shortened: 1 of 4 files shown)


alexandreLamarre:
Along with this change fixes conflicting alerting svc labels that meant we couldn't port-forward `svc/opni-monitoring-internal`

kralicky (approved):

PR #446, opened 2022-08-01
Filter duplicate filenames when searching for plugins with multiple patterns

### LINKED ISSUE #445: Plugin loader can load the same file more than once when given multiple search patterns

kralicky:
Fixes #445


dbason (approved):
LGTM

As discussed need to review the flaky timing test

PR #457, opened 2022-08-02
Add create cluster to opensearch management API

### LINKED ISSUE #422: Implement management API for logging plugin
Need an api for creating Opensearch clusters from the Gateway management UI.

Initial code will be behind a feature toggle until testing and all functions are complete

dbason:
Fixes #422


plugins/logging/main.go
```diff
@@ -2,20 +2,61 @@ package main
...
-	plugins.Serve(logging.Scheme(ctx))
-	ca()
+	for {
+		if err := runPlugin(ctx, ca); err != nil {
+			panic(err)
+		}
+	}
+}
+
+func runPlugin(ctx context.Context, cancel context.CancelFunc) error {
```
kralicky, line 28: When the gateway reloads, it will terminate and re-launch all plugins, so you might not need to handle it here.

packages/opni/opni/package.yaml
```diff
@@ -1,6 +1,6 @@
 
 url: local
-version: 0.5.4
+version: 0.5.5-dan-dev
```
kralicky, line 3: Intentional change?

kralicky (changes requested):

packages/opni/opni/package.yaml
dbason: Yes, but probably shouldn't be committed to main

plugins/logging/main.go
dbason: OK cool, I wasn't sure if it would leave the plugins running so I handled it here as a fallback.  I will drop it from here.

kralicky (approved):

PR #467, opened 2022-08-04
Ship traces to opensearch

### LINKED ISSUE #425: Configure tracing for dataprepper

dbason:
Fixes #425


pkg/util/waitctx/context.go
```diff
@@ -98,17 +98,55 @@ func (restrictive) Wait(ctx RestrictiveContext, notifyAfter ...time.Duration) {
...
+				fmt.Fprint(os.Stderr, chalk.Yellow.Color("\n=== WARNING: waiting longer than expected for context to cancel ===\n"+string(debug.Stack())+"\n"))
+
+			}
+		}
+	}
+	<-done
+	return
+}
+
+func (restrictive) WaitWWithTimeout(ctx RestrictiveContext, timeout time.Duration, notifyAfter ...time.Duration) {
```
kralicky, line 115: Unfortunately I think the only way to implement this safely is to panic in response to the timeout, which might not be what you want (or it might be).

kralicky (changes requested):

pkg/util/waitctx/context.go
dbason: So this was an interim solution.  With the Wait method blocking correctly it's introduced an issue with the integration tests.  It appears that there is something that is not ending correctly when the context it uses is cancelled.  This is just a stop gap until that bug can be fixed.

kralicky: If you try merging https://github.com/rancher/opni/tree/reload-bugfixes, does it work correctly? Several bugs were fixed here related to reload and waitctx not completing.

kralicky (approved):

PR #485, opened 2022-08-10
Opensearch template index

AmartC:
This PR adds logic to the Opensearch Updating Service by creating the templates index which stores the template, the template cluster ID as well as the most frequently occurring log message associated with that template. It receives this information through the Opni DRAIN service.

### DIFF
--- a/aiops/opni-opensearch-update-service/opensearch-update-service/app/main.py
+++ b/aiops/opni-opensearch-update-service/opensearch-update-service/app/main.py
@@ -23,10 +23,11 @@ script_for_anomaly = (
     "ctx._source.anomaly_predicted_count += 1; ctx._source.opnilog_anomaly = true;"
 )
 
-async def doc_generator(df):
+async def doc_generator(df, index, op_type="index"):
     main_doc_keywords = {"_op_type", "_index", "_id", "doc"}
-    df["_op_type"] = "update"
-    df["_index"] = "logs"
+    if op_type != "index":
+        df["_op_type"] = op_type
+    df["_index"] = index
     df.rename(columns={"log_id": "_id"}, inplace=True)
     for index, document in df.iterrows():
         doc_dict = document.to_dict()
@@ -37,8 +38,6 @@ async def doc_generator(df):
                 doc_dict["doc"][k] = doc_dict[k]
                 del doc_dict[k]
         yield doc_dict
-
-
 
 async def setup_es_connection():
     ES_ENDPOINT = os.environ["ES_ENDPOINT"]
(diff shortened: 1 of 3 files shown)


aiops/opni-opensearch-update-service/opensearch-update-service/app/main.py
```diff
@@ -122,6 +167,19 @@ async def update_logs(es, df):
...
+        exists = await es.indices.exists("templates")
+        if not exists:
+            await es.indices.create(index="templates")
+
+    except TransportError as exception:
+        logging.info(f"Error in es indices {exception}")
+        if exception.status_code == "N/A":
+            logging.info("Elasticsearch connection error")
+
+
```
dbason, line 182: This should be managed as part of the operator code we have for creating indices.  We should include an index template if the document in the index will be a standard format that we control.

As an example here is where we create some index templates - https://github.com/rancher/opni/blob/main/pkg/resources/opnicluster/elastic/indices/indices.go#L193-L219

And here is where we create an index:
https://github.com/rancher/opni/blob/main/pkg/resources/opnicluster/elastic/indices/indices.go#L262-L266

dbason (changes requested):

aiops/opni-opensearch-update-service/opensearch-update-service/app/main.py
```diff
@@ -23,10 +23,11 @@
     "ctx._source.anomaly_predicted_count += 1; ctx._source.opnilog_anomaly = true;"
 )
 
-async def doc_generator(df):
+async def doc_generator(df, index, op_type="index"):
     main_doc_keywords = {"_op_type", "_index", "_id", "doc"}
-    df["_op_type"] = "update"
-    df["_index"] = "logs"
+    if op_type != "index":
```
tybalex, line 28: unnecessary `if`

aiops/opni-opensearch-update-service/opensearch-update-service/app/main.py
```diff
@@ -23,10 +23,11 @@
     "ctx._source.anomaly_predicted_count += 1; ctx._source.opnilog_anomaly = true;"
 )
 
-async def doc_generator(df):
+async def doc_generator(df, index, op_type="index"):
```
tybalex, line 26: default op_type should be `update`.

tybalex (changes requested):
pls clean up debugging loggers.

aiops/opni-opensearch-update-service/opensearch-update-service/app/main.py
AmartC: By default, if _op_type isn't specified, it's index.

dbason (approved):

dbason (approved):

tybalex (approved):

tybalex (approved):

PR #543, opened 2022-09-04
New proto generate for log mgmt

dbason:


dbason:
@kralicky looks like the new generation is adding additional details into the atomicMessage field for the health check that the tests aren't expecting.

kralicky:
Pushed a change to fix this.

kralicky (approved):

PR #558, opened 2022-09-14
Fixes for logging mgmt

dbason:
Adds API to fetch storageclass names
Switches resource.Quantity to use strings due to problem with protojson marshalling.


plugins/logging/pkg/logging/admin.go
```diff
@@ -582,11 +637,57 @@ func convertDashboardsToProtobuf(dashboard opsterv1.DashboardsConfig) *loggingad
...
+				}
+			}
+
+			return resources
 		}(),
 	}
 }
+
+func replaceInArray[T comparable](array []T, old T, new T) []T {
+	newArray := make([]T, len(array))
```
kralicky, line 684: Should this be `make([]T, 0, len(array))`?

kralicky (changes requested):

kralicky (approved):

kralicky (approved):

PR #561, opened 2022-09-14
Add delete logging data function

dbason:
Switches to the new APIs.
Add NATs Jetstream cluster.
Delete logging data from opensearch.


plugins/logging/pkg/logging/admin.go
```diff
@@ -370,35 +411,115 @@ func (p *Plugin) convertProtobufToDashboards(
...
+	}
+
+	// Set sane transport timeouts
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.DialContext = (&net.Dialer{
+		Timeout: 5 * time.Second,
+	}).DialContext
+	transport.TLSHandshakeTimeout = 5 * time.Second
+	transport.TLSClientConfig = &tls.Config{
+		InsecureSkipVerify: true,
```
kralicky, line 460: Would it be doable to obtain the opensearch cacert here?

plugins/logging/pkg/logging/uninstall.go
```diff
@@ -2,19 +2,312 @@ package logging
...
+			backoff.WithMinInterval(5*time.Second),
+			backoff.WithMaxInterval(1*time.Minute),
+			backoff.WithMultiplier(1.1),
+		)
+		b := p.Start(ctx)
+	RETRY:
+		for {
+			select {
+			case <-b.Done():
+				ti.AddLogEntry(zapcore.WarnLevel, "Uninstall canceled, but time series data is still being deleted by Cortex")
```
kralicky, line 82: Update log message

plugins/logging/pkg/logging/uninstall.go
```diff
@@ -2,19 +2,312 @@ package logging
...
+	if err != nil {
+		return nil
+	}
+
+	if idExists {
+		entry, err := a.kv.Get(id)
+		if err != nil {
+			return nil
+		}
+		createNewJob = !(string(entry.Value()) == pendingValue)
```
kralicky, line 222: Perhaps `string(entry.Value()) != pendingValue` might be more clear :smile:

go.mod
```diff
@@ -121,6 +121,7 @@ require (
 	k8s.io/apimachinery v0.24.3
 	k8s.io/client-go v0.24.3
 	k8s.io/component-base v0.24.3
+	k8s.io/kubernetes v1.13.0
```
kralicky, line 124: there's a non-canonical api import somewhere here that is adding this as a dependency

kralicky (changes requested):

go.mod
dbason: Yeah it's a hangover from when I imported storage incorrectly.  I'll fix it up.

plugins/logging/pkg/logging/admin.go
dbason: Possibly, but its internal comms with a selfsigned cert so I was les concerned about it.

kralicky (approved):

PR #589, opened 2022-09-27
SLO dashboards, float precision, deleted rule masking & alerting mem leak fix

alexandreLamarre:


test/e2e/monitoring_test.go
```diff
@@ -231,7 +230,7 @@ var _ = Describe("Monitoring Test", Ordered, Label("e2e", "slow"), func() {
 				}
 			}
 			return fmt.Errorf("metrics capability not installed within time limit")
-		}, time.Second*60, time.Second*2).Should(Succeed())
```
kralicky, line 234: Are you observing this taking a long time? It should be very quick.

alexandreLamarre: I can do so more testing, but from memory its usually instant but there may have been times when its taken around 20 seconds

alexandreLamarre: The capability uninstall, with delete should be set 20 mins, must've accidentally set the capability install to that time

PR #592, opened 2022-09-28
Logging V2 plugin

dbason:


plugins/logging/pkg/gateway/stream.go
```diff
@@ -0,0 +1,17 @@
...
+	streamext "github.com/rancher/opni/pkg/plugins/apis/apiextensions/stream"
+	"github.com/rancher/opni/plugins/logging/pkg/apis/node"
+)
+
+func (p *Plugin) StreamServers() []streamext.Server {
+	return []streamext.Server{
+		{
+			Desc:              &node.NodeLoggingCapability_ServiceDesc,
+			Impl:              &p.logging,
+			RequireCapability: wellknown.CapabilityMetrics,
```
kralicky, line 14: Metrics -> Logging (although this field is currently unused, but it will be used later)

plugins/logging/pkg/backend/install.go
```diff
@@ -0,0 +1,131 @@
...
+	"google.golang.org/protobuf/types/known/emptypb"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	opensearchv1 "opensearch.opster.io/api/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (b *LoggingBackend) CanInstall(ctx context.Context, _ *emptypb.Empty) (*emptypb.Empty, error) {
```
kralicky, line 26: This is never called for agent v2 - this logic should be used, but abstracted away (see metrics ClusterDriver implementation)

plugins/logging/pkg/backend/install.go
```diff
@@ -0,0 +1,131 @@
...
+		Name:      b.OpensearchCluster.Name,
+		Namespace: b.OpensearchCluster.Namespace,
+	}, opensearchCluster); err != nil {
+		return nil, errors.ErrCheckOpensearchClusterFailed(err)
+	}
+
+	return &emptypb.Empty{}, nil
+}
+
+func (b *LoggingBackend) Install(ctx context.Context, req *capabilityv1.InstallRequest) (*capabilityv1.InstallResponse, error) {
```
kralicky, line 54: The install request API here should be modified slightly, to work like this:

1. Make a call to a ClusterDriver to check if the capability can be installed (using the CanInstall logic above)
2. If the IgnoreWarnings flag is not set, return the warning message as an error. Otherwise:
3. Make a call to the ClusterDriver to handle implementation-specific install logic. If it fails, you can choose to honor the IgnoreWarnings field or not, depending on if the system can reconcile itself later upon updating the opensearch cluster config. If continuing:
4. Add the capability name to the cluster's capability list.
5. Request a node sync
6. Return an install response with the relevant status and warning fields filled in. (for backwards compatibility purposes, returning nil here it will treat it as a success)

Note that your "ClusterDriver" equivalent interface will do more things than mine, since for metrics there are no extra steps I have to take when adding a cluster.

plugins/logging/pkg/backend/sync.go
```diff
@@ -0,0 +1,235 @@
...
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/timestamppb"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
```
kralicky, line 30: The LoggingBackend should be relatively generic - this is the intersection between the generic API (CapabilityBackend) and the implementation-specific API (NodeLoggingCapability). Both servers should for the most part forward requests to an implementation-specific "ClusterDriver" (or similar).

PR #621, opened 2022-10-05
Topology plugin base infrastructure

alexandreLamarre:
High Level Idea

image


alexandreLamarre:
@AmartC high level overview of the infrastructure implementation

topology_impl_overview

PR #641, opened 2022-10-09
Logging bugfixes

dbason:


plugins/logging/pkg/util/opensearch.go
```diff
@@ -0,0 +1,62 @@
+package util
+
+import (
+	"sync"
+
+	osclient "github.com/opensearch-project/opensearch-go"
+)
+
+type AsyncOpensearchClient struct {
```
kralicky, line 9: util.Initializer was designed to have a usable zero-value and lazy-initialize the mutex. Since you are creating it explicitly you should set this up with a constructor function instead.

```go
func NewAsyncOpensearchClient() *AsyncOpensearchClient {
  return &AsyncOpensearchClient{
    initCond: sync.NewCond(&sync.Mutex{}),
  }
}
```
Also, using the client as its own Locker is fine, but you won't be able to use the client itself concurrently (i'm guessing it is designed to allow this), and code using it would be prone to deadlocks.

kralicky (approved):

kralicky (approved):

PR #718, opened 2022-10-25
Training gateway plugin

AmartC:
This PR introduces the model training plugin for the gateway which will be used by the admin UI service to display workload deployments and log counts and launch model training jobs.


plugins/modeltraining/pkg/modeltraining/aggregation.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	a.ByCluster[bucket.Key.ClusterID].ByNamespace[bucket.Key.NamespaceName].ByDeployment[bucket.Key.DeploymentName] = &DeploymentLogCount{
+		DeploymentName: bucket.Key.DeploymentName,
+		Count:          bucket.DocCount,
+	}
+}
+
+func (s *ModelTrainingPlugin) run_aggregation() {
+	for {
+		request := map[string]interface{}{
```
kralicky, line 76: This should either be defined using a custom struct or a json string, nested `map[string]any` is kind of clunky.

plugins/modeltraining/pkg/modeltraining/aggregation.go
```diff
@@ -0,0 +1,169 @@
...
+			var result SearchResponse
+			if err := json.NewDecoder(res.Body).Decode(&result); err != nil {
+				log.Fatalf("Error parsing the response body: %s", err)
+			}
+			for _, b := range result.Aggregations.Bucket.Buckets {
+				resultAgg.Add(b)
+			}
+			after_key := result.Aggregations.Bucket.AfterKey
+			if after_key != nil {
+				((request["aggs"].(map[string]interface{})["bucket"]).(map[string]interface{})["composite"]).(map[string]interface{})["after"] = after_key
```
kralicky, line 158: Refactor this to read from the result struct

plugins/modeltraining/pkg/modeltraining/aggregation.go
```diff
@@ -0,0 +1,169 @@
...
+			}
+			res, err := s.osClient.Get().Search(
+				s.osClient.Get().Search.WithContext(context.Background()),
+				s.osClient.Get().Search.WithIndex("logs"),
+				s.osClient.Get().Search.WithBody(&buf),
+				s.osClient.Get().Search.WithTrackTotalHits(true),
+				s.osClient.Get().Search.WithPretty(),
+			)
+			if err != nil {
+				log.Fatalf("Error getting response: %s", err)
```
kralicky, line 146: This should probably not be a fatal error

plugins/modeltraining/pkg/modeltraining/aggregation.go
```diff
@@ -0,0 +1,169 @@
...
+					},
+				},
+			},
+		}
+		resultAgg := newAggregations()
+		for {
+			var buf bytes.Buffer
+			//var r map[string]interface{}
+			if err := json.NewEncoder(&buf).Encode(request); err != nil {
+				log.Fatalf("Error getting response: %s", err)
```
kralicky, line 135: This can be a fatal error (since it would be a bug) but the error message doesn't match.

plugins/modeltraining/pkg/modeltraining/aggregation.go
```diff
@@ -0,0 +1,169 @@
...
+				resultAgg.Add(b)
+			}
+			after_key := result.Aggregations.Bucket.AfterKey
+			if after_key != nil {
+				((request["aggs"].(map[string]interface{})["bucket"]).(map[string]interface{})["composite"]).(map[string]interface{})["after"] = after_key
+			} else {
+				break
+			}
+		}
+		aggregated_results, _ := json.Marshal(resultAgg)
```
kralicky, line 163: Handle error here

plugins/modeltraining/pkg/modeltraining/api.go
```diff
@@ -0,0 +1,139 @@
...
+
+	corev1 "github.com/rancher/opni/pkg/apis/core/v1"
+	util "github.com/rancher/opni/pkg/util/k8sutil"
+	modeltraining "github.com/rancher/opni/plugins/modeltraining/pkg/apis/modeltraining"
+	"google.golang.org/protobuf/types/known/emptypb"
+	k8scorev1 "k8s.io/api/core/v1"
+)
+
+func (c *ModelTrainingPlugin) TrainModel(ctx context.Context, in *modeltraining.WorkloadsList) (*corev1.Reference, error) {
+	var modelTrainingParameters = map[string]map[string][]string{}
```
kralicky, line 17: Use struct types instead of nested maps to model your data

plugins/modeltraining/pkg/modeltraining/api.go
```diff
@@ -0,0 +1,139 @@
...
+				parametersArray = append(parametersArray, &deployment_data)
+			}
+		}
+	}
+	trainingParameters.List = parametersArray
+	return &trainingParameters, nil
+}
+
+func (c *ModelTrainingPlugin) GpuPresentCluster(ctx context.Context, in *emptypb.Empty) (*modeltraining.GPUInfoList, error) {
+	client, err := util.NewK8sClient(util.ClientOptions{})
```
kralicky, line 115: Creating a new client every time this is called seems inefficient.

plugins/modeltraining/pkg/modeltraining/aggregation.go
```diff
@@ -0,0 +1,169 @@
...
+		bytes_aggregation := []byte(aggregated_results)
+		s.kv.Get().Put("aggregation", bytes_aggregation)
+		s.Logger.Info("Updated aggregation of deployments to Jetstream.")
+		time.Sleep(30 * time.Second)
```
kralicky, line 167: The interval at which this logic runs should be controlled by a ticker (`time.Ticker`) which can be canceled by a context. for example:
```go
t := time.NewTicker(interval)
for {
	select {
	case <-t.C:
    doStuff()
	case <-ctx.Done():
		t.Stop()
		return
	}
}
```

plugins/modeltraining/pkg/modeltraining/plugin.go
```diff
@@ -0,0 +1,95 @@
...
+	k8sClient      client.Client
+	osClient       future.Future[*opensearch.Client]
+	natsConnection future.Future[*nats.Conn]
+	kv             future.Future[nats.KeyValue]
+}
+
+func (s *ModelTrainingPlugin) UseManagementAPI(api managementv1.ManagementClient) {
+	lg := s.Logger
+	lg.Info("querying management API...")
+	var list *managementv1.APIExtensionInfoList
```
kralicky, line 37: Is this being used?

kralicky (changes requested):

kralicky:
Overall, need to ensure variable/function/type names follow the Go style guide https://go.dev/doc/effective_go#names

plugins/modeltraining/pkg/modeltraining/system.go
```diff
@@ -0,0 +1,118 @@
...
+				if i == 1 {
+					retryBackoff.Reset()
+				}
+				return retryBackoff.NextBackOff()
+			},
+		),
+	)
+}
+
+func getOpensearchCredentials(ctx context.Context) (username string, password string) {
```
dbason, line 49: You should be using the helper function from the operator
```
username, password, err := helpers.UsernameAndPassword(r.ctx, r.client, cluster)
	if err != nil {
		return err
	}
```
In this case cluster is the `opsterv1.OpenSearchCluster` object which you will need to fetch from kubernetes.

Avoid returning hardcoded credentials as this is masking errors that should probably be exposed as missing details, and it will also fall over when the user changes the default admin password.

dbason: Fetching the opensearch object is where you want to use your retried logic.  If you get an IsNotFound error you should retry, as this won't exist until the logging backend is created.

plugins/modeltraining/pkg/modeltraining/plugin.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+type ModelTrainingPlugin struct {
+	modeltraining.UnsafeModelTrainingServer
+	system.UnimplementedSystemPluginClient
+	ctx            context.Context
+	Logger         *zap.SugaredLogger
+	k8sClient      client.Client
+	osClient       future.Future[*opensearch.Client]
+	natsConnection future.Future[*nats.Conn]
```
dbason, line 30: You don't really need to store the natsConnection in state, just the KV as the connection is only used to create that KV

dbason (changes requested):

plugins/modeltraining/pkg/modeltraining/aggregation.go
dbason: An alternative could be to use gjson as a shortcut to maintaining structs - https://github.com/tidwall/gjson

plugins/modeltraining/pkg/modeltraining/aggregation.go
```diff
@@ -0,0 +1,169 @@
...
+				s.osClient.Get().Search.WithContext(context.Background()),
+				s.osClient.Get().Search.WithIndex("logs"),
+				s.osClient.Get().Search.WithBody(&buf),
+				s.osClient.Get().Search.WithTrackTotalHits(true),
+				s.osClient.Get().Search.WithPretty(),
+			)
+			if err != nil {
+				log.Fatalf("Error getting response: %s", err)
+			}
+			defer res.Body.Close()
```
dbason, line 148: If you're going to have a fatal error on the body not being valid json you need to check for
```
defer resp.Body.Close()
if resp.IsError() {
	return false, fmt.Errorf("failed to do the thing: %s", resp.String())
}

File: plugins/modeltraining/pkg/modeltraining/aggregation.go (hunk 2 above)
Comment (kralicky) [reply plugins/modeltraining/pkg/modeltraining/aggregation.go:158]: In this case I believe the structs are already in use here

File: plugins/modeltraining/pkg/modeltraining/aggregation.go [hunk 13]
```diff
@@ -0,0 +1,169 @@
...
+			var buf bytes.Buffer
+			//var r map[string]interface{}
+			if err := json.NewEncoder(&buf).Encode(request); err != nil {
+				log.Fatalf("Error getting response: %s", err)
+
+			}
+			res, err := s.osClient.Get().Search(
+				s.osClient.Get().Search.WithContext(context.Background()),
+				s.osClient.Get().Search.WithIndex("logs"),
+				s.osClient.Get().Search.WithBody(&buf),
```
dbason, line 141: Consider using the `JSONReader` util:
```
opensearchutil.NewJSONReader(request)
```
https://pkg.go.dev/github.com/opensearch-project/opensearch-go@v1.1.0/opensearchutil?utm_source=gopls#NewJSONReader

dbason (changes requested):

plugins/modeltraining/pkg/modeltraining/plugin.go
AmartC: Actually, the Nats connection is used in the api.go function to send over payload to the AI training controller service. 
Example: https://github.com/rancher/opni/blob/training-gateway-plugin/plugins/modeltraining/pkg/modeltraining/api.go#L42

plugins/modeltraining/pkg/modeltraining/api.go
```diff
@@ -0,0 +1,151 @@
...
+	ByDeployment map[string]*Deployment `json:",inline,omitEmpty"`
+}
+
+type Deployment struct {
+	DeploymentName string `json:",inline,omitEmpty"`
+}
+
+func (c *ModelTrainingPlugin) TrainModel(ctx context.Context, in *modeltraining.WorkloadsList) (*corev1.Reference, error) {
+	var modelTrainingParameters = map[string]map[string][]string{}
+	for idx := 0; idx < len(in.List); idx++ {
```
dbason, line 33: The correct way to enumerate a list in go is
```
for index, item := range in.List {}
```
For what you're doing you don't actually need the index either because you just want to operator on each item so you can do
```
for _, item := range in.List {}
```

dbason (changes requested):

plugins/modeltraining/pkg/modeltraining/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+	s.k8sClient.Set(client)
+	s.natsConnection.Set(nc)
+	s.kv.Set(keyValue)
+	osClient, err := s.newOpensearchConnection()
+	if err != nil {
+		lg.Error(err)
+	}
+	s.osClient.Set(osClient)
+	go s.runAggregation()
+
```
dbason, line 65: The errors in this function should probably be fatal.  If they don't get handled here they will cause nil pointer issues later in your code which can make it harder to trace back the problems

dbason (changes requested):

dbason (changes requested):

plugins/modeltraining/pkg/modeltraining/aggregation.go
```diff
@@ -0,0 +1,184 @@
...
+	ByDeployment map[string]*DeploymentLogCount `json:",inline,omitEmpty"`
+}
+
+type DeploymentLogCount struct {
+	DeploymentName string `json:"deployment_name"`
+	Count          int    `json:"doc_count"`
+}
+
+func newAggregations() *Aggregations {
+	return &Aggregations{ByCluster: make(map[string]*ClusterAggregation)}
```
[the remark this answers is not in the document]
AmartC: I ran it and nothing was changed.

plugins/modeltraining/pkg/modeltraining/aggregation.go
AmartC: https://github.com/rancher/opni/issues/724

dbason (approved):

plugins/modeltraining/pkg/modeltraining/system.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+
+	"github.com/cenkalti/backoff"
+	backoffv2 "github.com/lestrrat-go/backoff/v2"
```
kralicky, line 11: You're using two separate backoff libraries here, should probably pick one.

plugins/modeltraining/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/rancher/opni/pkg/plugins"
+	"github.com/rancher/opni/pkg/tracing"
+	"github.com/rancher/opni/pkg/util/waitctx"
+	modeltraining "github.com/rancher/opni/plugins/modeltraining/pkg/modeltraining"
+)
+
+func main() {
+	tracing.Configure("plugin_model_training")
```
kralicky, line 15: this should be `plugin_modeltraining`

kralicky (changes requested):

kralicky:
Also need to squash and reword similar commits

plugins/modeltraining/pkg/modeltraining/api.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	clusterAggregationResults, ok := resultsStorage.ByCluster[in.Id]
+	if !ok {
+		return &modeltraining.WorkloadsList{}, nil
+	}
+	workloadsList := modeltraining.WorkloadsList{}
+	workloadArray := make([]*modeltraining.WorkloadResponse, 0)
+	for namespaceName, deployments := range clusterAggregationResults.ByNamespace {
+		for deploymentName, count := range deployments.ByDeployment {
+			workloadAggregation := modeltraining.WorkloadResponse{ClusterId: in.Id, Namespace: namespaceName, Deployment: deploymentName, Count: int64(count.Count)}
```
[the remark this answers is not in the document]
AmartC: No, it is an integer.

plugins/modeltraining/pkg/modeltraining/api.go
```diff
@@ -0,0 +1,135 @@
...
+			workloadArray = append(workloadArray, &workloadAggregation)
+		}
+	}
+	workloadsList.List = workloadArray
+	return &workloadsList, nil
+}
+
+func (c *ModelTrainingPlugin) ModelStatus(ctx context.Context, in *emptypb.Empty) (*corev1.Reference, error) {
+	b := []byte("model_status")
+	msg, err := c.natsConnection.Get().Request("model_status", b, time.Minute)
```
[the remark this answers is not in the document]
AmartC: It needs to just send a byte array to the Nats subject.

plugins/modeltraining/pkg/modeltraining/api.go
```diff
@@ -0,0 +1,137 @@
...
+		allocatable := node.Status.Allocatable
+		for k, v := range capacity {
+			if strings.HasPrefix(string(k), "nvidia.com/gpu") {
+				allocation := allocatable[k]
+				gpuInfo := &modeltraining.GPUInfo{Name: string(k), Capacity: v.String(), Allocatable: allocation.String()}
+				gpuInfoArray = append(gpuInfoArray, gpuInfo)
+			}
+		}
+	}
+	returnedData.Items = gpuInfoArray
```
dbason, line 133: You're not using returnedData for anything so it should be removed.

plugins/modeltraining/pkg/modeltraining/system.go
dbason: This is copied from me, they do disparate things.  The cenkalti package allow us to return an exponentially increasing duration (and importantly reset it) for the nats custom reconnect duration.

The lestrrat-go essentially provides a ticker interface with exponential time increases

dbason (changes requested):

plugins/modeltraining/pkg/apis/modeltraining/modeltraining.proto
```diff
@@ -0,0 +1,57 @@
...
+    option (google.api.http) = {
+      get: "/modelTrainingParameters"
+    };
+  }
+  rpc GpuPresentCluster(google.protobuf.Empty) returns (GPUInfoList) {
+    option (google.api.http) = {
+      get: "/gpuPresentCluster"
+    };
+  }
+}
```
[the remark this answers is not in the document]
AmartC: We shouldn't be passing in the cluster ID since this is only for the central Opni cluster.

dbason (approved):

plugins/modeltraining/pkg/modeltraining/plugin.go
```diff
@@ -0,0 +1,85 @@
...
+	lg := s.Logger
+	nc, err := newNatsConnection()
+	if err != nil {
+		lg.Fatal(err)
+	}
+	mgr, err := nc.JetStream()
+	if err != nil {
+		lg.Fatal(err)
+	}
+	keyValue, err := mgr.CreateKeyValue(&nats.KeyValueConfig{
```
kralicky, line 43: If the kv store already exists, will this error? Might need to check if it exists first.

kralicky (changes requested):

kralicky (dismissed):

PR #740, opened 2022-10-28
Add jetstream storage backend; various bugfixes and test improvements

### LINKED ISSUE #726: Jetstream Storage Backend

kralicky:
Closes #726


alexandreLamarre (approved):
LGTM

alexandreLamarre (approved):

pkg/storage/jetstream/jetstream.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	options.apply(opts...)
+
+	lg := logger.New(logger.WithLogLevel(zap.WarnLevel)).Named("jetstream")
+
+	nkeyOpt, err := nats.NkeyOptionFromSeed(conf.NkeySeedPath)
+	if err != nil {
+		return nil, err
+	}
+	nc, err := nats.Connect(conf.Endpoint, nkeyOpt, nats.MaxReconnects(-1))
```
dbason, line 69: May want to add a custom disconnect handler e.g:
```
nats.DisconnectErrHandler(
	func(nc *nats.Conn, err error) {
		lg.With(
			"err", err,
		).Warn("nats disconnected")
	},
),
```

dbason: A `nats.CustomReconnectDelay(func(int) time.Duration)` might also be an idea to prevent spamming the nats cluster if something happens

pkg/storage/jetstream/jetstream.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return store, nil
+}
+
+func (s *JetStreamStore) upsertBucket(name string) nats.KeyValue {
+	bucketName := fmt.Sprintf("%s-%s", s.BucketPrefix, name)
+	kv, err := s.js.KeyValue(bucketName)
+	if err != nil {
+		if errors.Is(err, nats.ErrBucketNotFound) {
+			kv, err = s.js.CreateKeyValue(&nats.KeyValueConfig{
```
dbason, line 99: `CreateKeyValue` is idempotent.  If the bucket already exists it will throw an error if the config is different to the existing bucket.  It may be simpler to just call create every time.

kralicky: It is not documented to be idempotent, and there is a warning above `nats.KeyValueManager` stating it is experimental and the functionality may change. Without an API guarantee, it is probably safer  to initialize the bucket this way, in case they change the implementation details later.

pkg/storage/jetstream/jetstream.go
kralicky: :+1: also added RetryOnFailedConnect(true) which looks like it is needed to enable retries on the initial connection.

alexandreLamarre (approved):

dbason (approved):

dbason (approved):

PR #749, opened 2022-11-01
automation for publishing chart updates to charts-repo branches

kralicky:


ci.cue
```diff
@@ -35,8 +37,9 @@ dagger.#Plan & {
 			HELM_OCI_REPO:          string | *"ghcr.io/rancher"
 			DASHBOARDS_VERSION:     string | *"1.3.3"
 			OPENSEARCH_VERSION:     string | *"1.3.3"
-			PLUGIN_VERSION:         string | *"0.6.0"
-			PLUGIN_PUBLISH:         string | *"0.6.0"
+			PLUGIN_VERSION:         string | *opniVersion
+			PLUGIN_PUBLISH:         string | *opniVersion
```
dbason, line 41: These should probably be left as is.  We don't always tag a new release on the opensearch/dashboards plugins when we cut RCs (although I guess we could start)

dbason (dismissed):

PR #760, opened 2022-11-02
Update opensearch plugin config

### LINKED ISSUE #758: Enable both ingest plugin in a future release
Now we have 2 plugins:
the Logging plugin: normalize the `time` and `log` field from incoming logs, which is needed for the dashboard. It also detects Json objects in the logs.
the AIOps plugin: do some extra preprocessing that needed for the AI services + sending logs to Nats.

Ideally, if the Logging backend is enabled but AIOps not enabled , only the Logging plugin should be installed: and if AIOps got enabled, both plugin should be installed in the pipeline.

dbason:
Fixes #758

### DIFF
--- a/ci.cue
+++ b/ci.cue
@@ -504,6 +504,17 @@ dagger.#Plan & {
 							]
 						}
 					},
+					docker.#Run & {
+						command: {
+							name: "opensearch-plugin"
+							args: [
+								"-s",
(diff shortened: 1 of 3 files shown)


kralicky (approved):

tybalex (approved):
approved

PR #793, opened 2022-11-10
rework dashboard backend using gin router with proxy support

kralicky:


codyrancher (dismissed):
Looks fine from my perspective

PR #831, opened 2022-11-16
downstream capability alerts & cortex backend alerts

alexandreLamarre:
- Generic time series
The generic time series allow to persist long lived data (state of the condition) with a ttl so that we can query the state of conditions over a long period of time, while also minimizing the information stored for each condition.
When we determine that some incoming state for a condition is different that its last known state, then we store this new state. For our purposes we restrict the time series state to only storing 3 things `health` : is the condition specified by the user met?, `firing` is the condition is a state where the user should be notified?, `timestamp` : when was the last state change?
- Generic evaluators:
Generic evaluators have 5 components, `metadata`, `context`, `storage`, `state` and `hooks`. `metadata` is responsible for owning metadata related to the alert type / user configuration for use during evaluation. `context` is responsible for holding references to the current evaluation context, which can be cancelled if it encouters a non-recoverable error or when the user changes a configuration that would impact the evaluation, and also responsible for holding any parent or stream contexts that are necessary for determining if the evaluator should exit early or perform some other action. `storage` holds references to the alerting storage node, so that it can persist last known states, time series, stream identifiers, channels etc...  `state` is the current in memory state to allow for quick lookups for transitioning between healthy, unhealthy, firing, retry and other states the evaluator could encounter. `hooks` contain closures that are run on specific state changes, namely `healthOnMessage` -> evaluate health on each incoming message, `triggerHook` -> what actions to undertake when the alert fires and `resolveHook` -> what actions to undertake when the alert resolves. In the future there will also be finalizer hooks for some evaluators that should exit in an expected manner, for example dispatching a notification when some process is done.
- Durable consumers
Durable consumers back some ingress streams for the alerting plugin in order to be able to replay a certain amount of messages for the persistent streams. This maintains cohesion of states between alarms without having to share data as well as improve fault tolerance if the alerting plugin is unavailable. The alerting plugin sets up watchers and hooks use by durable consumers; they are set up after the Plugin acquires the management client (see `system.go`/`UseWatchers` method.) and this is where the durable consumers should also be set. 
- Downstream capability alerts
Uses the durable consumer backed stream from agent healthstatus messages & the generic evaluator/time series to allow the user to track downstream agent states like `Pending`, `Failure` over a set amount of time to be able to track failures in the downstream agents.
- Cortex backend alerts
Changed some of the error handling in the cortexadmin plugin; when certain components are unavailable due to invalid configs or "rare" issues they can return `ECONNREFUSED`/`DNSERROR` which we now return as internal server errors when querying the cortex components so that the alerting plugin can know that this is indeed an invalid state as a result of some underlying issue.
Cortex backend alerts use a heartbeat watcher to fetch the information needed from the cortexadmin status API so there is no need to back these streams using a durable consumer.


plugins/metrics/pkg/apis/cortexadmin/status.proto
```diff
@@ -12,6 +14,7 @@ message CortexStatus {
   StoreGatewayStatus storeGateway = 6;
   QueryFrontendStatus queryFrontend = 7;
   QuerierStatus querier = 8;
+  google.protobuf.Timestamp ts = 9;
```
kralicky, line 17: Suggest renaming to `timestamp`

pkg/apis/alerting/v1/alerting.proto
```diff
@@ -79,11 +79,13 @@ enum AlertType {
   CPU_SATURATION = 2;
   MEMORY_SATURATION = 3;
   FS_SATURATION = 4;
+  DOWNSTREAM_CAPABILTIY = 5;
```
kralicky, line 82: use TitleCase for proto enums (unless for compatibility reasons)

Tiltfile
```diff
@@ -1,10 +1,10 @@
 load('ext://min_k8s_version', 'min_k8s_version')
-load('ext://cert_manager', 'deploy_cert_manager')
+# load('ext://cert_manager', 'deploy_cert_manager')
 load('ext://namespace', 'namespace_create')
 
 set_team('52cc75cc-c4ed-462f-8ea7-a543d398a381')
 
-version = '0.6.0'
+version = '0.6.2-rc1'
```
kralicky, line 7: This should probably match the version in ci.cue in main

pkg/apis/alerting/v1/extensions.go
```diff
@@ -218,3 +231,53 @@ func (a *AlertCondition) SetClusterId(clusterId *corev1.Reference) error {
 	}
 	return shared.WithInternalServerErrorf("AlertCondition could not find its clusterId")
 }
+
+func mergeLabels(ms ...map[string]string) map[string]string {
```
kralicky, line 235: This is `lo.Assign` if you want to use that instead

pkg/apis/alerting/v1/alerting.proto
```diff
@@ -281,6 +289,34 @@ message AlertConditionPrometheusQuery{
   google.protobuf.Duration for = 3;
 }
 
+message AlertConditionDownstreamCapability{
+  core.Reference clusterId = 1;
+  repeated string capabilityState = 2;
+  google.protobuf.Duration for = 3;
```
kralicky, line 295: `for` is too ambiguous of a name here IMO

pkg/alerting/shared/streams.go
```diff
@@ -9,21 +9,65 @@ import (
```
kralicky, line 9: Any cortex/alertmanager-specific code should be confined to the plugins as much as possible; if you can, keep only the string constants here.

kralicky (changes requested):

pkg/alerting/shared/streams.go
alexandreLamarre: Done

Tiltfile
alexandreLamarre: Done

plugins/metrics/pkg/apis/cortexadmin/status.proto
alexandreLamarre: Done

kralicky (approved):

kralicky (approved):

kralicky (approved):

PR #849, opened 2022-11-19
force efficient remount of config map in alertmanager pod

alexandreLamarre:


plugins/alerting/pkg/alerting/drivers/alerting_driver_dynamic.go
```diff
@@ -194,9 +200,125 @@ func (a *AlertingManager) Update(ctx context.Context, conf *alertops.AlertingCon
...
+	var numWorkerReplicas *int32
+	zero := int32(0)
+	err = a.k8sClient.Get(ctx, client.ObjectKeyFromObject(workerSvcData), workerSvcData)
+	if k8serrors.IsNotFound(err) {
+		numWorkerReplicas = &zero
+	} else if numWorkerReplicas == nil {
+		numWorkerReplicas = &zero
+	} else {
+		numWorkerReplicas = workerSvcData.Spec.Replicas
+	}
```
kralicky, line 237: This logic doesn't look quite right - also the default value for Replicas is 1 if it's nil

plugins/alerting/pkg/alerting/drivers/alerting_driver_dynamic.go
```diff
@@ -194,9 +200,125 @@ func (a *AlertingManager) Update(ctx context.Context, conf *alertops.AlertingCon
...
+				)
+				if err == nil {
+					if cmp.IsEmpty() {
+						return status.Error(codes.FailedPrecondition, "no changes to apply")
+					}
+				}
+				return a.k8sClient.Update(ctx, clone)
+			})
+			if err != nil {
+				lg.Error(err)
```
kralicky, line 316: Should probably do something with this error, maybe just have more descriptive logging if you want this to be best-effort and not return an error

kralicky (changes requested):

kralicky (approved):

PR #876, opened 2022-11-30
Model status plugin

AmartC:
This PR extends the training gateway plugin by including an endpoint for receiving post requests that give the current status of a model for training.


plugins/aiops/pkg/apis/modeltraining/modeltraining.proto
```diff
@@ -13,6 +13,12 @@ service ModelTraining {
       body: "*"
     };
   }
+  rpc PostModelTrainingStatus(ModelTrainingStatistics) returns (ModelTrainingResponse) {
+    option (google.api.http) = {
+      post: "/model/statistics"
```
dbason, line 18: As this is calling a PUT on the kv (i.e. it is idempotently updating information) it should also be a PUT in the REST API, not a POST.

dbason (changes requested):
One comment, and also it looks like you have introduced an additional modeltraining.pb.go file

kralicky (approved):

dbason (approved):

PR #887, opened 2022-12-08
Have training plugin reset statistics for next model

AmartC:
This PR resets the training statistics for a model when a new training job is submitted.

### DIFF
--- a/plugins/aiops/pkg/gateway/modeltraining.go
+++ b/plugins/aiops/pkg/gateway/modeltraining.go
@@ -34,7 +34,7 @@ func (p *AIOpsPlugin) TrainModel(ctx context.Context, in *modeltraining.ModelTra
 	}
 	msg, err := p.natsConnection.Get().Request("train_model", jsonParameters, time.Minute)
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Failed to train model: %v", err)
+		return nil, status.Errorf(codes.Unavailable, "Failed to train model: %v", err)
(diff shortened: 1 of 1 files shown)


kralicky (approved):

plugins/aiops/pkg/gateway/modeltraining.go
```diff
@@ -36,6 +36,10 @@ func (p *AIOpsPlugin) TrainModel(ctx context.Context, in *modeltraining.ModelTra
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to train model: %v", err)
 	}
+	_, err = p.PutModelTrainingStatus(ctx, &modeltraining.ModelTrainingStatistics{RemainingTime: 3600})
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to put model training status: %v", err)
```
kralicky, line 41: Set the code to `Unavailable` which indicates the call can be retried (if the server is temporarily down for example)

kralicky (changes requested):

plugins/aiops/pkg/gateway/modeltraining.go
kralicky: sorry I put this comment on the wrong line, it should be on line 37

AmartC: Done

kralicky (approved):

PR #912, opened 2022-12-15
Custom grpc gateway error handling

### LINKED ISSUE #893: Improve grpc error handling
As of right now grpc can emit an input parsing error but provides very little context about what field caused the error.

I'd like to see additional context added to the error messages so that I can immediately determine what area of the input is causing the issue instead of playing a trial and error game hunting for it.

kralicky:
This PR improves the consistency of grpc-to-http error message conversion for all core management apis by removing the extra json formatting on the native grpc messages (which is not useful to the ui) and making sure error codes are consistent.

Closes #893


dbason:
I've restarted the test flake so need to wait for that to pass

codyrancher (approved):
Already tested and verified that the frontend error handling works with this change.

dbason (approved):

PR #923, opened 2022-12-16
alertingv2 spec draft

alexandreLamarre:


alexandreLamarre:
Closing, since the need for a second version of the api is stale.

PR #931, opened 2022-12-19
Add state tracking to admin creation

### LINKED ISSUE #930: Track state of admin user creation in persistent store
We only want to create the Opensearch admin user once, so it doesn't get recreated if the end user intentionally deletes it, however we should track the state of the create process so we can restart it if it never completed and the gateway restarts

dbason:
This will ensure that if the gateway is restarted before the opensearch admin user is successfully created the creation of the admin user will be picked up again.

Fixes #930

### DIFF
--- a/pkg/test/environment.go
+++ b/pkg/test/environment.go
@@ -23,6 +23,8 @@ import (
 	"text/template"
 	"time"
 
+	natsserver "github.com/nats-io/nats-server/v2/server"
+	natstest "github.com/nats-io/nats-server/v2/test"
 	"github.com/nats-io/nats.go"
 	"github.com/nats-io/nkeys"
 	"github.com/rancher/opni/pkg/alerting/metrics"
@@ -134,6 +136,8 @@ type Environment struct {
 
 	gatewayConfig *v1beta1.GatewayConfig
 	k8sEnv        *envtest.Environment
+
+	embeddedJS *natsserver.Server
 
 	Processes struct {
 		Etcd      future.Future[*os.Process]
@@ -503,6 +507,25 @@ func (e *Environment) Context() context.Context {
 	return e.ctx
 }
 
+func (e *Environment) StartEmbeddedJetstream() (*nats.Conn, error) {
+	ports, err := freeport.GetFreePorts(1)
+	if err != nil {
+		return nil, err
+	}
+
+	opts := natstest.DefaultTestOptions
+	opts.Port = ports[0]
+
+	e.embeddedJS = natstest.RunServer(&opts)
+	e.embeddedJS.EnableJetStream(nil)
+	if !e.embeddedJS.ReadyForConnections(2 * time.Second) {
+		return nil, errors.New("starting nats server: timeout")
+	}
+
+	sUrl := fmt.Sprintf("nats://127.0.0.1:%d", ports[0])
+	return nats.Connect(sUrl)
+}
+
 func (e *Environment) StartK8s() (*rest.Config, *runtime.Scheme, error) {
 	e.initCtx()
 	e.Processes.APIServer = future.New[*os.Process]()
@@ -631,6 +654,9 @@ func (e *Environment) Stop() error {
 	}
 	if e.k8sEnv != nil {
 		e.k8sEnv.Stop()
+	}
+	if e.embeddedJS != nil {
+		e.embeddedJS.Shutdown()
 	}
 	if e.mockCtrl != nil {
 		e.mockCtrl.Finish()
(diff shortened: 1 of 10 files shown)


pkg/test/environment.go
```diff
@@ -503,6 +507,25 @@ func (e *Environment) Context() context.Context {
 	return e.ctx
 }
 
+func (e *Environment) StartEmbeddedJetstream() (*nats.Conn, error) {
```
kralicky, line 510: Let's add this as an option to `Start()` which would replace the default subprocess jetstream server (you can also keep this function). We should also be able to obtain the port for the embedded server the same way as the external jetstream.

Additionally, try running all the tests using the embedded jetstream to see which ones work and which don't. This could potentially speed up some tests.

pkg/test/environment.go
```diff
@@ -135,6 +137,8 @@ type Environment struct {
 	gatewayConfig *v1beta1.GatewayConfig
 	k8sEnv        *envtest.Environment
 
+	embeddedJS *natsserver.Server
```
kralicky, line 140: Could the lifetime of this server be controlled by ctx directly?

kralicky (changes requested):

pkg/test/environment.go
dbason: There doesn't seem to be any wrappers that take a context so we would need to do this ourselves if we wanted to.

pkg/test/environment.go
dbason: Let's add that as a separate task.  I don't want to couple the implementation of this feature with additional test modifications.

pkg/test/environment.go
```diff
@@ -503,6 +507,25 @@ func (e *Environment) Context() context.Context {
 	return e.ctx
 }
 
+func (e *Environment) StartEmbeddedJetstream() (*nats.Conn, error) {
+	ports, err := freeport.GetFreePorts(1)
```
alexandreLamarre, line 511: Can you add an associated port in the Environment ports struct  & an environment variable for this port?
Also could you delete the hackier version of starting jetstream in the testenv : https://github.com/rancher/opni/blob/a49ecd410a51ebf95fa80025ff0bbabd63ff955e/pkg/test/environment.go#L650

pkg/test/environment.go
dbason: #933 created to track this

pkg/test/environment.go
dbason: As discussed in Joe's comment I would prefer to not tie that work to the implementation of this feature.  I also think there is value in running a fatter Jetstream server for the test.Env as it also sets up nkey and other things that are used in integration tests.  The embedded server is a minimal server just for unit tests.

plugins/logging/pkg/gateway/admin_v2.go
```diff
@@ -1008,6 +1012,7 @@ FETCH:
 }
 
 func (m *LoggingManagerV2) generateAdminPassword(cluster *loggingv1beta1.OpniOpensearch) (password []byte, retErr error) {
+	ctx := context.TODO()
```
alexandreLamarre, line 1015: Since this is the plugin-side external reconciler, does it makes sense to pass in the Logging plugin ctx to the `LoggingManagerV2` and use this as a context here?

dbason: possibly but I think the use of contexts, and the passing in of them needs careful review.  This is intended as a stopgap pending that review which is why it's `context.TODO()` instead of `context.Background()`

plugins/logging/pkg/gateway/admin_v2.go
```diff
@@ -1019,10 +1024,43 @@ func (m *LoggingManagerV2) generateAdminPassword(cluster *loggingv1beta1.OpniOpe
...
+		password = secret.Data["password"]
+		return
+	}
+
 	return
 }
 
+func (m *LoggingManagerV2) createInitialAdmin() error {
+	k8sOpensearchCluster := &loggingv1beta1.OpniOpensearch{}
+	ctx := context.TODO()
```
alexandreLamarre, line 1045: Same consideration as previous context.TODO()

pkg/test/environment.go
kralicky: I agree, adjusting existing tests can be a separate task. The first part of my comment is still relevant though.

pkg/test/environment.go
kralicky: In that case we should do it ourselves by adding a waitctx action.

PR #946, opened 2023-01-05
ListRules as a command line diagnostic tool

alexandreLamarre:
- list rules fixes for cortex HA mode
- multi cluster list support
- Improved GRPC api :
  - accepts filters
  - returns protobuf message instead of raw bytes
- CLI for listing & filtering rules


plugins/metrics/pkg/cortex/api_types.go
```diff
@@ -0,0 +1,124 @@
...
+	"regexp"
+
+	"golang.org/x/exp/slices"
+)
+
+func xOR(expr1, expr2 bool) bool {
+	return expr1 != expr2
+}
+
+type listRuleFilter struct {
```
joshmeranda, line 13: Do we really need a new type here? Seems like there's some coed duplication in `NewListRuleFilter` and the validation function `func (l *ListRulesRequest) Validate() error` in `plugins/metrics/pkg/apis/cortexadmin/extensions.go`.

Can we refactor this to a couple methods instead?

```go
func MatchesCluster(request *cortexadmin.ListRulesRequest, clusterId string) bool {
	if *request.RequestAll {
		return true
	}

	// we store each cluster rules in a file with the cluster id
	// and when we search invalid we search specifically those that do not match existing clusters
	return xOR(*request.ListInvalid, slices.Contains(request.ClusterId, clusterId))
}
```

LGTM otherwise

alexandreLamarre: Yep

joshmeranda (approved):
LGTM

PR #952, opened 2023-01-06
Cortex Rule CRUD through CLI

alexandreLamarre:
Decided to limit LoadRules & DeleteRule api to a single cluster to limit unintended side effects from the caller


joshmeranda (approved):
LGTM

joshmeranda (approved):

Issue #790, opened 2022-11-10
Cortex rule invalid bucket-index.json.gz


alexandreLamarre:
rules/bucket-index.json.gz contains a null block 
```
version	1
blocks	null
block_deletion_marks	[]
updated_at	1668098050
```
Seems to only happen on monitoring startup/ metrics capability startup

- completely soft locks cortex functionality  
- deleting all the cortex pods doesn't resolve the issue
- deleting the invalid bucket-index.json.gz doesn't resolve the issue
- deleting the rules data in the bucket doesn't resolve the issue

alexandreLamarre:
Hitting this again in 0.8rc5

alexandreLamarre:
For reference a normal `bucket-index.json.gz` should look like:
```
[...]

PR #953, opened 2023-01-06
Fix cortex null block soft-block on functionality

alexandreLamarre:
Fixes https://github.com/rancher/opni/issues/790


alexandreLamarre:
PR is stale, closing.

Issue #966, opened 2023-01-23
Create human readable cluster names


joshmeranda:
Currently when selecting which agent cluster whose logs to view, the user must select from a list of agent IDs:

![logging-clsuter-ids](https://user-images.githubusercontent.com/43455015/214072152-e1dcba25-5e7d-4ff4-a3e4-c5ecb5a838ea.png)

Ideally the user would be able to specify more readable names to select from.

joshmeranda:
closed in favor of tracking issue: https://github.com/rancher/opni/issues/1057

PR #969, opened 2023-01-24
OEP Add Human Readable Agent Names for Log Analytics

joshmeranda:
Related to https://github.com/rancher/opni/issues/966


sanjay920 (changes requested):
Clusters can already be renamed. There should be a way to do this in the CLI already. @codyrancher @kralicky can you chime in here?
The UI now gives clusters a name if the user doesn't input one when adding an agent
The goal of this proposal should be to use the cluster name, which should be human readable, in opensearch dashboards. Opni Grafana is already doing this so maybe @kralicky can share some details on that
For cluster names to make their way to opensearch, we need a separate index that Opni creates upon install and a trigger that writes to this index any time:
* a new cluster is added
* an existing cluster name is changed

I'm also thinking about how names should be treated when an agent is removed. I think it would be good to add `(inactive)` as a postfix to clusters that are no longer sending data to opni. @dbason

codyrancher:
> . Opni Grafana is already doing this so maybe @kralicky can share some details on that

I'm not sure what more I can add here. What you've stated about the UI is accurate. It's fine if you want to put the cluster name in a separate index but it will force me to do a join so if there's no extra cost it could go into the existing docs if you wanted.

`(inactive)` is a presentation detail. We should probably just mark the state of the agent in the document in case in the future we want to mark things like disconnected or something.

sanjay920:
I agree we should have cluster name in the document so that user's can search and filter on it. We'd have to update all documents with any updates to cluster names

enhancements/logging/20230122-readable-logging-index-name.md
```diff
@@ -0,0 +1,44 @@
...
+   1. replace the name label
+   2. remove the pre-existing name label
+2. Create a new alias pointing to the existing index for the agent
+
+For this implementation to make the most sense, the names passed to `opni cluster rename <cluser> <name>` should be unique. While the implementation would still work with duplicate names, the user would see the logs from both agents in the dashboard when the user selects the index.
+
+## Acceptance criteria:
+
+- [ ] Running `opni cluster rename <cluster> <name>` should add a new index called `<name>` to the index dropdown list
+
```
dbason, line 31: Are we going to have a separate index per cluster?  Would this not be better stored as a single index with a document for each cluster?

dbason:
We could add the cluster name to the documents when we index in the same way that we add the cluster ID now (i.e. an arbitrary field added).  This can be done using either dataprepper or fluentd.  The downside to this is reindexing the documents if/when the cluster name changes.  That is potentially going to be very expensive depending on the amount of historical data.  If cluster names are immutable this is less of a problem, but if they are not doing a join might be the cheaper way to go.

joshmeranda:
@dbason I initially thought we were already storing our cluster data in different indexes so I figured it would be simpler and cheaper to add aliases to those existing indexes rather than having to iterate over all the data again.

But if we are already identifying the cluster id in the documents, unless there is some cheap way of iterating over all those documents we probably don't want to deal with that. I think mutable cluster names should be mutable to allow for users to rename clusters if they change purpose / domain and need to conform to a naming convention, otherwise they'd probably need to reinstall the agent to change names. They also wouldn't need to know the name at install time, which would make it easier to play around with opni before fully adopting it.

I'm thinking we might want to just handle this on the gateway side with some mapping of ids to their user facing names, and expose that mapping as a service which the UI can ping for a list of those cluster names, and when needed. This wouldn't do much for users trying to query using open search since it wouldn't know of those mappings, but that might be out of scope here.

dbason:
So doing some research on joins in Opensearch I think they add a level of complexity we don't necessarily want to deal with.  In particular when indexing logs we'd need to ensure we had the id of the parent document:
>When indexing child documents, you have to specify the routing query parameter because parent and child documents in the same relation have to be indexed on the same shard. Each child document refers to its parent’s ID in the parent field .I think this would make indexing logs burdensome.

If we stored the cluster name in the event of the cluster name changing we would need to do an update_by_query operation on the existing documents.  This may be somewhat expensive but we can run it as a background task.

I do still think we should consider having the cluster_name in a separate table and manually joining them in the UI.  We could ease the issue with the discover page by presenting a list of clusters in the plugin and linking out to the discover page with the id filter prefilled.

@codyrancher @sanjay920 thoughts on the last suggestion?

dbason:
It should probably be done in a generic fashion.  We could probably use the [task](https://github.com/rancher/opni/tree/main/pkg/task) machinery to do this and have a rename task similar to how we have an uninstall task - @kralicky thoughts?  We could then use the Opensearch tasks API in the same way that we do for delete_by_query [here](https://github.com/rancher/opni/blob/main/plugins/logging/pkg/opensearchdata/delete.go)

kralicky:
I would need more info about how the opensearch dashboards work under the hood - in particular, do we have the ability to query the management API to obtain cluster names like we do in the admin dashboard?

joshmeranda:
Is there anything else we might want to add to this cluster name index? Following @codyrancher 's suggestion we can add a cluster status field so we can mark delete clusters as inactive. Probably isn't something we want to include in this OEP, just getting this on the radar

dbason:
> I would need more info about how the opensearch dashboards work under the hood - in particular, do we have the ability to query the management API to obtain cluster names like we do in the admin dashboard?

Why is that relevant?  I believe what we would need to is add an API function to rename a downstream cluster (analogous to uninstalling a capability from a cluster).  This would then need to call the rename task for any installed capabilities.  That way responsibility would be devolved to each individual plugin.

I think the status is probably a sensible solution but will require more thought.  We have been discussing a soft delete in the admin API, and this should be part of the consideration of that.  That said if a user decides to uninstall a capability without deleting the data I don't think presenting that data to them in the datastore is a problem.

dbason (approved):

sanjay920 (approved):

PR #972, opened 2023-01-24
Interopability with Prometheus AlertManager OEP

alexandreLamarre:


enhancements/alerting/20230124-routing-metrics-backend.md
```diff
@@ -0,0 +1,144 @@
...
+## Supporting documents:
+
+## Dependencies:
+
+- https://github.com/rancher/opni/pull/942
+
+## Risks and contingencies:
+
+| Risk | Contingency |
+| Some AlertManager endpoints may require premium licenses for functionality | Request license through SUSE |
```
kralicky, line 132: Check to see if alertmanager/cortex/etc tests these already, if so we can skip

kralicky (dismissed):

enhancements/alerting/20230124-routing-metrics-backend.md
alexandreLamarre: The issue is that adding a config that violates structural or runtime assumptions may break all the runtime alertmanager functionality - with very little indication of what is wrong.

My philosophy is we shouldn't let the users hang themselves, if possible.

One other option is to buffer the error and debug logs from within the AlertManager process and group these logs by endpoint, then expose these via an API.  But that will be stateful and still doesn't help when the process exits abnormally.

The core requirement of Opni-Alerting is not presupposing any knowledge about AlertManager or Opni backends, I'm still leaning towards manually testing them and having as excellent as possible input validation to prevent these issues. Thoughts?

dbason (approved):

enhancements/alerting/20230124-routing-metrics-backend.md
kralicky: That sounds reasonable to me. :+1:

kralicky (approved):

PR #975, opened 2023-01-25
OEP Opensearch over gRPC

dbason:


enhancements/logging/20230125-proxy-opensearch-over-grpc.md
```diff
@@ -0,0 +1,50 @@
...
+## Supporting documents: 
+N/A
+
+## Dependencies: 
+None; all dependencies have been completed.
+
+## Risks and contingencies: 
+| Risk | Contingency|
+|---|---|
+| Shipping logs over gRPC causes significant performance issues in either Agent or Gateway | Investigate alternatives using compression or refactor of the proxy process.  Additional scaling may be required|
```
kralicky, line 42: I would add that in the event we encounter performance issues, we should try changing the grpc server/client options [here](https://github.com/rancher/opni/blob/d7eb01c0c677d71c592aeaa21ff6a936625c053b/pkg/gateway/grpc.go#L77-L93) and [here](https://github.com/rancher/opni/blob/d7eb01c0c677d71c592aeaa21ff6a936625c053b/pkg/clients/gateway_client.go#L125-L136) - it is possible these are not the most optimal combination of options for the messages we are sending.

enhancements/logging/20230125-proxy-opensearch-over-grpc.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Impact: 
+The change will largely be transparent to end users.
+
+Potential drawbacks are as follows:
+Resilience - agent becomes a critical single point of failure (although this is arguably already the case)
+Performance - a transparent proxy should not involve any signficant performance issues, however it is more data running through the agent which will have an impact.
+
+## Implementation details: 
+For the downstream cluster a new streaming plugin with an http interface will be created. Queries received on the http will be proxied over gRPC to the Gateway. The gateway will forward the query to Opensearch and return the response.
```
kralicky, line 22: Do you mean this will be in a separate plugin than the logging plugin?

dbason: The logging plugin currently doesn't have a streaming plugin attached to it so this will be net new.

kralicky: The logging plugin uses the stream for capability backend sync, would you add the new grpc services there?

dbason:
This OEP has been deprecated.

PR #996, opened 2023-01-30
Write up OTEL logging proposal

dbason:


kralicky (dismissed):
:shipit:

enhancements/logging/20230130-use-otel-for-logging
```diff
@@ -0,0 +1,82 @@
...
+## Benefits: 
+* Improve maintainbilty my moving preprocessing out of an internal java project to a separate service
+* Simplify sending logs via the Agent as OTLP is a gRPC format
+* Align with the Open Telemetry project for logging
+* We don't expose the Opensearch API outside of the central cluster so we never have to transmit credentials, or worry about escalated access to the API.
+
+## Impact: 
+* Remove dependence on the Banzaicloud operator. Fluentbit and Fluentd would be removed. However this means we will need to build our own log aggregation controller.
+* This also means we will diverge from rancher-logging in its current form.
+* We will need a migration path to deprecate the ingest plugin.
```
sanjay920, line 18: What would the migration plan be for users upgrading from the old logging capability to this one? This needs to be included here at a high level at minimum

sanjay920 (changes requested):
enhancements/logging/20230130-use-otel-for-logging should have the .md extension

enhancements/logging/20230130-use-otel-for-logging
dbason: @sanjay920 I've added details to the Implementation section and Acceptance Criteria that detail the migration path

dbason:
Re the http configuration I think that this will be better served as a separate enhancement in the future.  The config refresh will be the tricky thing here.  I'm not sure adding an additional listening endpoint for configuration is the best move, and I don't want to require that problem to be solved before we implement this, which has immediate benefits.

kralicky:
FWIW the http endpoint for agents already exists (it's currently used for prometheus remote write), as well as the http api extensions mechanism.

dbason:
I mean the endpoint to allow the otel collector to be refreshed, or otherwise adding in a signal handler (although I guess we could run this as a sidecar).  Either way I think it's best left as a future enhancement.  I have created a discussion for it [here](https://github.com/rancher/opni/discussions/1016)

sanjay920 (approved):

kralicky (approved):

PR #1012, opened 2023-02-01
Add mage target for vale styles

alexandreLamarre:
Managing vale styles between git work trees and different computers becomes tedious, so introducing so simple automation here

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -3,11 +3,14 @@
 package main
 
 import (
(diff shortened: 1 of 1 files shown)


alexandreLamarre:
Built `vale` from source and the issue I was seeing with `vale sync` on my linux distribution has been fixed in their main branch

PR #1014, opened 2023-02-02
Prometheus Remote Read ETL tool

joshmeranda:
OEP: https://github.com/rancher/opni/pull/970


kralicky (changes requested):
Looks good so far. Couple of things to do before we can merge:
- There are a lot of todos, you should attempt to resolve as many as possible (all the ones that aren't blocked by other things)
- There are a couple of style related comments, do a once over through all the new code and make sure code style stays consistent.

joshmeranda:
Sounds good, adding a few more tests I thought of while  writing docs, so I'll add those, and start addressing your comments. Thanks!

plugins/metrics/pkg/agent/http.go
```diff
@@ -132,3 +133,66 @@ func (s *HttpServer) handlePushRequest(c *gin.Context) {
 		c.Status(http.StatusServiceUnavailable)
 	}
 }
+
+//func (s *HttpServer) handleRemoteReadStart(c *gin.Context) {
```
alexandreLamarre, line 137: Do these commented out functions serve a purpose?

joshmeranda: nope, I definitely just commented those out when I was still tinkering and forgot about them. I'll delete them

joshmeranda: nope, I definitely just commented those out when I was still tinkering and forgot about them. I'll delete them now

plugins/metrics/pkg/agent/node.go
```diff
@@ -149,13 +171,66 @@ func (m *MetricsNode) GetHealth(_ context.Context, _ *emptypb.Empty) (*corev1.He
...
+
+func (m *MetricsNode) GetTargetStatus(ctx context.Context, request *remoteread.TargetStatusRequest) (*remoteread.TargetStatus, error) {
+	m.targetRunnerMu.RLock()
+	defer m.targetRunnerMu.RUnlock()
+
+	return m.targetRunner.GetStatus(request.Meta.Name)
+}
+
+func (m *MetricsNode) Discover(ctx context.Context, request *remoteread.DiscoveryRequest) (*remoteread.DiscoveryResponse, error) {
+	driver, err := drivers.GetNodeDriver("external-operator")
```
kralicky, line 206: You should add the `Discover` method to the driver interface, feel free to refactor as necessary (it'll probably look weird but in this case it's okay for the driver interface to be a collection of unrelated methods). Maybe also rename to `Discover{RemoteReadTargets|Targets|Prometheuses|etc}`.

Adding the method will also require an implementation in the testenv driver, which will definitely be useful for integration tests.

plugins/metrics/pkg/agent/node.go
```diff
@@ -149,13 +171,66 @@ func (m *MetricsNode) GetHealth(_ context.Context, _ *emptypb.Empty) (*corev1.He
...
+}
+
 func (m *MetricsNode) doSync(ctx context.Context) {
 	m.logger.Debug("syncing metrics node")
 	m.nodeClientMu.RLock()
 	defer m.nodeClientMu.RUnlock()
 
 	if m.nodeClient == nil {
-		m.conditions.Set(health.CondConfigSync, health.StatusPending, "no client, skipping sync")
+		m.conditions.Set(health.CondConfigSync, health.StatusPending, "no promClient, skipping sync")
```
kralicky, line 233: Is it possible to get here? Looks like doSync is only called if nodeClient is not nil in the first place.

plugins/metrics/pkg/apis/remoteread/client.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+}
+
+func WithDialOptions(options ...grpc.DialOption) ClientOption {
+	return func(opt *ClientOptions) {
+		opt.dialOptions = append(opt.dialOptions, options...)
+	}
+}
+
+func NewClient(ctx waitctx.PermissiveContext, opts ...ClientOption) (RemoteReadGatewayClient, error) {
```
kralicky, line 36: Suggest renaming to convey that this is specifically for the gateway client

plugins/metrics/pkg/backend/metrics.go
```diff
@@ -408,3 +420,318 @@ func (m *MetricsBackend) UninstallCluster(ctx context.Context, in *emptypb.Empty
...
+func (m *MetricsBackend) AddTarget(_ context.Context, request *remoteread.TargetAddRequest) (*emptypb.Empty, error) {
+	m.WaitForInit()
+
+	m.remoteReadTargetMu.Lock()
+	defer m.remoteReadTargetMu.Unlock()
+
+	targetId := getIdFromTargetMeta(request.Target.Meta)
+
+	if _, found := m.remoteReadTargets[targetId]; found {
+		return nil, targetAlreadyExistsError(targetId)
```
kralicky, line 447: Generally, errors returned from grpc handler functions should be of the form `status.Error(codes.XXX, ...)`, otherwise the grpc status code gets set to Unknown which makes it harder to handle errors on the client side.
```suggestion
		return nil, status.Errorf(codes.NotFound, "target not found: %s", id)
```

plugins/metrics/pkg/cortex/api.go
```diff
@@ -25,6 +25,7 @@ type forwarders struct {
 	QueryFrontend gin.HandlerFunc
 	Alertmanager  gin.HandlerFunc
 	Ruler         gin.HandlerFunc
+	RemoteRead    gin.HandlerFunc
```
kralicky, line 28: Unused?

kralicky (changes requested):
Might also need to rebase

pkg/opni/util/render_cortex.go
```diff
@@ -202,3 +203,57 @@ func servicesByName[T interface {
...
 	return services
 }
+
+func RenderTargetList(list *remoteread.TargetList) string {
+	writer := table.NewWriter()
+	writer.SetStyle(table.StyleColoredDark)
+	writer.AppendHeader(table.Row{"CLUSTER", "NAME", "ENDPOINT", "LAST READ", "STATE", "MESSAGE"})
+
+	for _, target := range list.Targets {
+		if target == nil {
```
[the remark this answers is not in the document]
joshmeranda: you're right, when I added this I was getting weird nil entries in the received  `TargetList`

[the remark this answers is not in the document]
kralicky: If you are still getting that issue, it's likely a bug on the server side

plugins/metrics/pkg/cortex/api.go
joshmeranda: yep not used, I'm gonna do another once over to catch anything that I forget about when I was doing the initial integration and didn't know where things went

plugins/metrics/pkg/agent/node.go
joshmeranda: I think the only change I made here was accidentally chaning the `client` -> `promClient`.

Regardless I think in [MetricsNode.SetNodeClient](https://github.com/rancher/opni/blob/ff5c37f4522a74ad086bf86208f53fd876d78238/plugins/metrics/pkg/agent/node.go#L95-L102) its possible to pass a `nil` client. The only place its called is with a non-nil client, but its *possible*. It should be safe to remove the check, but I'd rather leave it in or move the check into `SetNodeClient`.

plugins/metrics/pkg/agent/node.go
```diff
@@ -208,13 +208,7 @@ func (m *MetricsNode) Discover(ctx context.Context, request *remoteread.Discover
...
 	}
 
-	promDriver, ok := driver.(*drivers.ExternalPromOperatorDriver)
-	if !ok {
-		m.logger.Errorf("found wrong driver type '%T'", driver)
-		return nil, fmt.Errorf("could not discover Prometheus instances")
-	}
-
-	entries, err := promDriver.Discover(*request.Namespace)
+	entries, err := driver.DiscoverPrometheuses(ctx, *request.Namespace)
```
kralicky, line 211: The driver here still needs to be dependency-injected. I suggest handling it somewhere in [here](https://github.com/rancher/opni/blob/main/plugins/metrics/pkg/agent/plugin.go#L34-L54) maybe by passing the driver to NewMetricsNode.

kralicky (approved):

kralicky (dismissed):

Issue #1020, opened 2023-02-03
Grpc unary caching interceptor


alexandreLamarre:
Implement a generic caching interceptor for API entity caching, to be used form unary requests within api extension calls.

alexandreLamarre:
In review, rolling over to sprint 4

PR #1029, opened 2023-02-06
GRPC caching interceptors

alexandreLamarre:
Addresses #1020


pkg/caching/caching.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"time"
+
+	"google.golang.org/protobuf/proto"
+
+	cache "github.com/patrickmn/go-cache"
+	"github.com/rancher/opni/pkg/storage"
+)
+
+type InMemoryEntityCache struct {
```
kralicky, line 12: This probably needs a memory limit as well.

kralicky (approved):

kralicky (changes requested):
wrong button

pkg/util/grpc.go
```diff
@@ -105,3 +143,219 @@ func StatusCode(err error) codes.Code {
...
+		clientMd, ok := metadata.FromOutgoingContext(ctx)
+		// forcibly disable caching if cache-control header is set to no-cache
+		if ok && noCache(clientMd, g.cacheType) {
+			return invoker(ctx, method, req, reply, cc, opts...)
+		}
+
+		// always try to
+		if !ok || shouldLookup(clientMd) {
+			// check if we have a cached value
+			lookupKey = g.hash(method, req.(proto.Message))
```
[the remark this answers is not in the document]
alexandreLamarre: Done

pkg/util/grpc.go
```diff
@@ -105,3 +143,219 @@ func StatusCode(err error) codes.Code {
...
+		md = metadata.New(map[string]string{})
+	}
+	md.Append(GrpcCacheStatusHeader(CacheTypeClient), GrpcCacheMiss)
+	md.Append(GrpcCacheControlHeader(CacheTypeClient), GrpcMaxAge(ttl))
+	md.Append(GrpcCacheControlHeader(CacheTypeClient), GrpcMustRevalidateCache)
+	grpc.SetHeader(ctx, md)
+}
+
+func (g *GrpcClientEntityCacher) hash(method string, req proto.Message) string {
+	return HashStrings([]string{method, string(Must(protojson.Marshal(req)))})
```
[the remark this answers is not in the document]
alexandreLamarre: Using wireframe now

pkg/util/grpc.go
```diff
@@ -105,3 +143,219 @@ func StatusCode(err error) codes.Code {
...
+	return fmt.Sprintf("%d", HashString(buf.String()))
+}
+
+// string hashes a string to a unique hashcode.
+//
+// crc32 returns a uint32, but for our use we need
+// a non negative integer. Here we cast to an integer
+// and invert it if the result is negative.
+func HashString(s string) int {
+	v := int(crc32.ChecksumIEEE([]byte(s)))
```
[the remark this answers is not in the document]
alexandreLamarre: looked into xxh3, seems applicable due to low overhead & speed

pkg/caching/caching.go
alexandreLamarre: using ccache LRU cache with memory limits & eviction policies now

pkg/util/grpc.go
```diff
@@ -105,3 +143,231 @@ func StatusCode(err error) codes.Code {
...
+		md = metadata.New(map[string]string{})
+	}
+	md.Append(GrpcCacheStatusHeader(CacheTypeClient), GrpcCacheMiss)
+	md.Append(GrpcCacheControlHeader(CacheTypeClient), GrpcMaxAge(ttl))
+	md.Append(GrpcCacheControlHeader(CacheTypeClient), GrpcMustRevalidateCache)
+	grpc.SetHeader(ctx, md)
+}
+
+func (g *GrpcClientEntityCacher) hash(method string, req proto.Message) string {
+	return HashStrings([]string{method, string(Must(proto.Marshal(req)))})
```
kralicky, line 209: Instead of proto.Marshal you need to do
```go
proto.MarshalOptions{
  Deterministic: true,
}.Marshal(req)
```

kralicky (changes requested):

pkg/util/grpc.go
alexandreLamarre: Done

kralicky (dismissed):

joshmeranda (approved):

dbason (approved):

pkg/util/grpc.go
```diff
@@ -45,6 +83,272 @@ func StatusCode(err error) codes.Code {
...
+
+func NewClientGrpcEntityCacher(cache storage.GrpcTtlCache) *GrpcClientEntityCacher {
+	return &GrpcClientEntityCacher{
+		GrpcTtlCache: cache,
+		cacheType:    CacheTypeClient,
+	}
+}
+
+// both the server and client agree on a signature that identifies the RPC that should be cached
+func (g *GrpcClientEntityCacher) sign(signature string) metadata.MD {
```
kralicky, line 282: I would use different terminology here, "signature" refers to a very specific kind of thing.

pkg/util/grpc.go
```diff
@@ -45,6 +83,272 @@ func StatusCode(err error) codes.Code {
...
+}
+
+type GrpcClientEntityCacher struct {
+	storage.GrpcTtlCache
+	cacheType string
+}
+
+var _ GrpcCachingInterceptor = (*GrpcClientEntityCacher)(nil)
+
+func NewClientGrpcEntityCacher(cache storage.GrpcTtlCache) *GrpcClientEntityCacher {
```
kralicky, line 274: Naming doesn't match (EntityCache/TtlCache)

kralicky (changes requested):

pkg/util/grpc.go
```diff
@@ -45,6 +83,272 @@ func StatusCode(err error) codes.Code {
...
+			append(opts,
+				grpc.Header(&cacheStatus),
+			)...,
+		)
+		if err != nil {
+			return err
+		}
+		if isCacheMiss(cacheStatus, g.cacheType) {
+			if shouldGrpcCache(clientMd, g.cacheType, method) {
+				storeKey = g.key(method, req.(proto.Message), ignoreCustomKeys(clientMd, g.cacheType))
```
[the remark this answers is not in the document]
alexandreLamarre: switched all proto.Message casts to use that utility method

pkg/management/server.go
```diff
@@ -144,7 +144,9 @@ func NewServer(
 		grpc.Creds(insecure.NewCredentials()),
 		grpc.UnknownServiceHandler(unknownServiceHandler(director)),
 		grpc.ChainStreamInterceptor(otelgrpc.StreamServerInterceptor()),
-		grpc.ChainUnaryInterceptor(otelgrpc.UnaryServerInterceptor()),
+		grpc.ChainUnaryInterceptor(
+			util.NewClientGrpcTtlCacher(nil).UnaryServerInterceptor(),
```
kralicky, line 148: Is this arg supposed to be nil?

pkg/caching/caching.go
```diff
@@ -3,11 +3,19 @@ package caching
 import (
 	"time"
 
+	"google.golang.org/protobuf/proto"
+	"k8s.io/apimachinery/pkg/api/resource"
```
kralicky, line 7: Don't import k8s.io from here

pkg/management/server.go
alexandreLamarre: yeah; we bundle both interceptors into one interface. Since the caching implementation is entirely client-side in these interceptors, when we only require a ServerInterceptor declaration we can omit the caching implementation.

Could add comment or implement a no-op cache to make this clearer.

pkg/caching/caching.go
```diff
@@ -39,17 +46,55 @@ func (i InMemoryHttpTtlCache) Delete(key string) {
...
+	maxAge time.Duration
+	// expired is a channel that is used to notify the cache
+	// that an item has to be delete
+}
+
+func (i InMemoryGrpcTtlCache) MaxAge() time.Duration {
+	return i.maxAge
+}
+
+func (i InMemoryGrpcTtlCache) Get(key string) (req proto.Message, ok bool) {
```
kralicky, line 71: These args named `req` should probably be named `resp` instead (also below in Set)

kralicky:
Should update to ccache v3

alexandreLamarre:
we had also had a discussion offline where the concern that hash collisions might cause unpredictable behaviour, so we are opting to only have the grpc caching interceptor cache when the req message implements CacheKeyer. 

This will also eliminate the concern towards returning a different return type from the expected one (in case of a very unfortunate collision)

pkg/caching/interceptors.go
```diff
@@ -0,0 +1,325 @@
...
+)
+
+func init() {
+	RegisterProtoType(&emptypb.Empty{}, func(_ protoreflect.ProtoMessage) string {
+		return "empty"
+	})
+}
+
+func protoId(msg protoreflect.ProtoMessage) string {
+	return string(msg.ProtoReflect().Descriptor().FullName().Name())
```
kralicky, line 61: Use the FullName here instead of the short name

pkg/caching/interceptors.go
```diff
@@ -0,0 +1,325 @@
...
+// - from a known map of registered proto messages
+func key(method string, req protoreflect.ProtoMessage) string {
+	val := reflect.Indirect(reflect.ValueOf(req))
+	typ := val.Type() // get the type of the underlying value
+	if reflect.PtrTo(typ).Implements(reflect.TypeOf((*CacheKeyer)(nil)).Elem()) {
+		return fmt.Sprintf("%s/%s", method, req.(CacheKeyer).CacheKey())
+	}
+	if keyFunc, ok := knownCacheableTypes[protoId((req))]; ok {
+		return fmt.Sprintf("%s/%s", method, keyFunc(req))
+	}
```
kralicky, line 232: looks like you can move this section to the top of the function

kralicky (changes requested):

pkg/caching/interceptors.go
alexandreLamarre: yep, good catch

pkg/apis/management/v1/extensions.go
```diff
@@ -15,3 +19,22 @@ func (cl *CapabilityList) Names() []string {
...
+	if len(m.MatchLabels.MatchExpressions) == 0 &&
+		len(m.MatchLabels.MatchLabels) == 0 &&
+		m.MatchOptions == 0 {
+		return "all"
+	}
+	key := m.MatchOptions.String()
+	for _, l := range m.MatchLabels.GetMatchExpressions() {
+		key += l.Key + l.Operator + strings.Join(l.Values, ",")
+	}
+	for k, v := range m.MatchLabels.GetMatchLabels() {
```
kralicky, line 36: this is a map, you will need to sort the keys

kralicky (approved):

PR #1031, opened 2023-02-07
Dependency upgrades

kralicky:
Note: Had to vendor code from an existing fork of cortex-tools, since it was very out-of-date and wasn't going to update easily


dbason (changes requested):
As discussed we also need to update the CRDs in `/config/agent-crd/bases`

dbason (approved):

alexandreLamarre (approved):

PR #1036, opened 2023-02-07
Implement OTEL gRPC handlers

### LINKED ISSUE #1032: OTLP gRPC added to logging plugins
Logs should be able to be transmitted from the downstream cluster to the central preprocessing service using OTLP.

Acceptance Criteria:

- [x] OTLP Server/Client architecture registered in Logging plugins
~- [ ] OTLP messages able to be forwarded over Gateway stream~

dbason:
Fixes #1032


plugins/logging/pkg/util/generic.go
```diff
@@ -0,0 +1,45 @@
...
+		if !alwaysSet {
+			return
+		}
+		c.Client = setter()
+		return
+	}
+	c.Client = setter()
+	c.initialized = true
+	c.initCond.Broadcast()
+}
```
kralicky, line 45: This logic doesn't really make sense in the context of AsyncClient as a whole. Perhaps this needs to be separate functions or separate types.

plugins/logging/pkg/util/generic.go
```diff
@@ -0,0 +1,54 @@
...
+	c.set = true
+	c.initCond.Broadcast()
+}
+
+// SetClient will always update the client regardless of its previous confition.
+func (c *AsyncClient[T]) SetClient(client T) {
+	c.initCond.L.Lock()
+	defer c.initCond.L.Unlock()
+	c.Client = client
+	c.set = true
```
kralicky, line 53: If c.set was previously false, you might need to call Broadcast here

plugins/logging/pkg/util/otel.go
```diff
@@ -0,0 +1,117 @@
...
+		b := expBackoff.Start(ctx)
+
+		for {
+			select {
+			case <-b.Done():
+				f.opts.lg.Warn("plugin context cancelled before gRPC client created")
+				return nil
+			case <-b.Next():
+				conn, err := grpc.Dial(
+					f.opts.collectorAddressOverride,
```
kralicky, line 95: Probably need an option to add custom `grpc.DialOption`s when creating the otel forwarder

plugins/logging/pkg/util/generic.go
dbason: Set client is meant to be called synchronously as opposed to background init.  The idea being that nothing should be waiting on the init condition when using set.

kralicky: If SetClient is incompatible with WaitForInit, i guess it should be a separate type? Otherwise it will appear to be a mistake since you're locking the condition variable mutex

Issue #1021, opened 2023-02-03
HTTP transport caching 


alexandreLamarre:
Implement a `http.RoundTripper` transport for http requests delegated through gateway/plugin http requests.

PR #1059, opened 2023-02-10
Implement http caching transport

alexandreLamarre:
Addresses #1021


alexandreLamarre:
Reopening in another PR, since drone is stuck on an old base branch

PR #1064, opened 2023-02-14
Add controllers for OTEL log scraper

dbason:


pkg/resources/labels.go
```diff
@@ -11,6 +11,7 @@ const (
 	OpniClusterID        = "opni.io/cluster-id"
 	OpniBootstrapToken   = "opni.io/bootstrap-token"
 	OpniInferenceType    = "opni.io/inference-type"
+	OpniConfigHash       = "opni.io/config-hash"
```
kralicky, line 14: Would you mind updating the gateway controller to use this annotation too? 
https://github.com/rancher/opni/blob/1ea59ec854d50bfe16dae6df61412ea6c13f93d3/pkg/resources/gateway/gateway.go#L71-L73

kralicky (approved):

PR #1067, opened 2023-02-14
Misc Operator Bugfixes

kralicky:


controllers/core_monitoring_controller.go
```diff
@@ -60,5 +65,30 @@ func (r *CoreMonitoringReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Watches(
+			&source.Kind{Type: &corev1beta1.Gateway{}},
+			handler.EnqueueRequestsFromMapFunc(r.findGateway),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
+		).
 		Complete(r)
 }
+
+func (r *CoreMonitoringReconciler) findGateway(o client.Object) []ctrl.Request {
+	monitoringClusters := &corev1beta1.MonitoringClusterList{}
```
dbason, line 77: If we're looking up `MonitoringClusters` wouldn't this be better of called `findMonitoringClusters` ?

dbason (changes requested):

controllers/core_monitoring_controller.go
kralicky: good catch

dbason (approved):

PR #1081, opened 2023-02-16
Logging & Alerting integration OEP

alexandreLamarre:


dbason (commented):
I think in the first instance it might be better off to strongly curate this more.  Standard alerting for logging is when error rate goes above a threshold, so I think we can put some wrappers around this.  Perhaps an alert for cluster status would be another one we could do.  Let's aim to have a sync up about it next week.

sanjay920 (approved):

dbason (approved):

Issue #974, opened 2023-01-25
Internal Datasource Alarms do not evaluate properly


alexandreLamarre:
After a long amount of time has passed while internal conditions should fire the batching mechanism (32) can lead to alarms evaluating to false if the the first and last messages of each batch meet `condition==false`, when there are multiple batches of (32) being processed

alexandreLamarre:
The root cause seems to be receiving a backlog of messages (pre-installation of alerting) which get batched in multiple groups of 32. I think an easy enough fix (logically) is to iterate & consume all batches until none are left in the subscriber component of the internal datasource evaluation.

Issue #1098, opened 2023-02-20
ListMessages API


alexandreLamarre:
- [x] Gateway API to read messages stored in cache
  - [x] Serve message LFU caches via an HTTP API
  - [x] Iterate & aggregate over the AlertManager "clientset" to get all messages
  - [x] Distinguish between "Alarm" and "Plain" messages

PR #1103, opened 2023-02-21
Persist alerting messages & List messages API

### LINKED ISSUE #1097: Alerting Cluster LFU Cache for messaging
Part of [Alerting Messaging System OEP](https://github.com/rancher/opni/blob/main/enhancements/alerting/20230124-messaging-system.md)

Opni-Alerting would like to provide a way to store messages even if there are no endpoints configured to receive those messages.

- [x]  LRU cache with keys based on priority and message frequency, turning it into an LFU cache
- [x] Distinguish between "Alarm" and "Plain" messages
- [x] Serve the messaging cache in the Opni server embedded in the AlertManager process

alexandreLamarre:
Closes #1097  #1098 #1125

- [X] Finalizer in the opni router points to the opni embedded server `HandleWebhook` endpoint inside the alertmanager process
- [X] The embedded server now handles caching messages based on priority approximately derived from `(severity, frequency)` pairs
  - [X] We maintain separate caches for message type `Alarm` and `Notification`
  - [X] For each cache we maitain a sub-cache for each severity
  - [X] The embedded server scales the importance of frequency linearly with the message severity and vice-versa the importance of "recent-ness" scales linearly with decreasing message severity

After talking to Cody,  it makes more sense for alarm messages to be correlated with particular incidents, where as notifications should be presented by occurrence.

From the original design we segregate ListMessages API into two APIs:
- [x] `ListAlarmMessages` : extracted from the cache by `(id, startTimestamp, endTimestamp)` to be correlated to particular incidents
- [x] `ListNotificationMessages` : extracted from the cache by their priority ordering based on `(severity,frequency)` pairs, but returned in sorted timestamp order
- [x] Incidents aren't available for any Prometheus based alarms, due to some existing limitations, so looking at introducing any prometheus based alarms into the timeline in this PR as well


alexandreLamarre:
Also closes https://github.com/rancher/opni/issues/974 .

We now only resolve internal alarms only if stream messages are consistently healthy during an evaluation cycle.

pkg/alerting/drivers/backend/api.go
```diff
@@ -183,6 +211,15 @@ func WithPostResolveAlertBody(conditionId string, labels, annotations map[string
 	}
 }
 
+func WithPostProto(req proto.Message) AlertManagerApiOption {
+	return func(o *AlertManagerApiOptions) {
+		bytes, err := json.Marshal(req)
```
dbason, line 216: should this instead be `protojson.Marshal` ?

pkg/alerting/drivers/routing/routing.go
```diff
@@ -74,7 +76,7 @@ func NewDefaultOpniRoutingWithOverrideHook(hook string) OpniRouting {
 }
 
 func NewDefaultOpniRouting() OpniRouting {
-	return NewOpniRouterV1("http://localhost:3000")
+	return NewOpniRouterV1(fmt.Sprintf("http://localhost:3000%s", shared.AlertingDefaultHookName))
```
dbason, line 79: Might want to put some belts and braces here to confirm that this constructs a valid URL

pkg/alerting/templates/funcs.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	timestamp := v * 1e9
+	if timestamp > math.MaxInt64 || timestamp < math.MinInt64 {
+		return nil, fmt.Errorf("%v cannot be represented as a nanoseconds timestamp since it overflows int64", v)
+	}
+	t := model.TimeFromUnixNano(int64(timestamp)).Time().UTC()
+	return &t, nil
+}
+
+func convertToFloat(i interface{}) (float64, error) {
```
dbason, line 26: Nit: I think @kralicky thinks we should use the `any` alias rather than empty interface.  I do agree that `any` communicates the intent better, but I'm not married to enforcing this.

pkg/apis/alerting/v1/extensions.go
```diff
@@ -1,30 +1,71 @@
...
+	NotificationPropertyClusterId = "opni_clusterId"
 	// Property that specifies how to classify the notification according to golden signal
-	NotificationPropertyGoldenSignal = "opni.io/goldenSignal"
+	NotificationPropertyGoldenSignal = "opni_goldenSignal"
 	// Property that specifies the severity of the notification. Severity impacts how quickly the
 	// notification is dispatched & repeated, as well as how long to persist it.
-	NotificationPropertySeverity = "opni.io/severity"
+	NotificationPropertySeverity = "opni_severity"
+	// Property that is used to correlate messages to particular incidents
+	NotificationPropertyFingerprint = "opni_fingerprint"
```
dbason, line 68: Out of interest why are we changing the format of labels from the k8s standard label?  Is this something to do with prometheus parsing?

alexandreLamarre: yep [prometheus alertmanager labels](https://prometheus.io/docs/alerting/latest/configuration/#configuration-file) don't accept alot of characters ` <labelname>: a string matching the regular expression [a-zA-Z_][a-zA-Z0-9_]* `.

pkg/alerting/drivers/routing/routing.go
alexandreLamarre: It eventually translates down to a `url.Url` inside the alertmanager config and is validated at that time, but yeah for posterity we should validate here as well

dbason (approved):

pkg/apis/alerting/v1/extensions.go
```diff
@@ -92,12 +142,38 @@ func (n *Notification) GetRoutingGoldenSignal() string {
...
+	res := map[string]string{
+		shared.OpniHeaderAnnotations:      a.header(),
+		shared.OpniBodyAnnotations:        a.body(),
+		shared.OpniClusterAnnotation:      a.GetClusterId().GetId(),
+		shared.OpniAlarmNameAnnotation:    a.GetName(),
+		shared.OpniGoldenSignalAnnotation: a.GetRoutingGoldenSignal(),
+	}
+	if IsMetricsCondition(a) {
+		res[NotificationPropertyFingerprint] = fmt.Sprintf(
+			"{{ \"ALERTS_FOR_STATE{opni_uuid=\\\"%s\\\"} OR vector(0)\" | query | first | value",
```
kralicky, line 168: Use raw string literals here instead

pkg/apis/alerting/v1/extensions.go
```diff
@@ -355,3 +417,35 @@ func (r *ListRoutingRelationshipsResponse) GetInvolvedConditions(endpointId stri
...
 }
+func (l *ListAlertConditionRequest) FilterFunc() func(*AlertCondition, int) bool {
+	return func(item *AlertCondition, _ int) bool {
+		if len(l.Clusters) != 0 {
+			if !slices.Contains(l.Clusters, item.GetClusterId().Id) {
+				return false
+			}
+		}
+		if len(l.Labels) != 0 {
+			if len(lo.Intersect(l.Labels, item.Labels)) != len(l.Labels) {
```
kralicky, line 428: This is going to allocate memory for a new map, I'd do this check by hand instead

pkg/apis/alerting/v1/extensions.go
```diff
@@ -355,3 +417,35 @@ func (r *ListRoutingRelationshipsResponse) GetInvolvedConditions(endpointId stri
...
+				return false
+			}
+		}
+		if len(l.Severities) != 0 {
+			if !slices.Contains(l.Severities, item.Severity) {
+				return false
+			}
+		}
+		if len(l.Severities) != 0 {
```
kralicky, line 437: possible bug here, should this be checking l.AlertTypes?

pkg/apis/alerting/v1/validate.go
```diff
@@ -505,5 +513,57 @@ func (n *Notification) Validate() error {
 			return validation.Error("if specifying a cluster id property, it must be set")
 		}
 	}
+	if _, ok := n.Properties[NotificationPropertyOpniUuid]; !ok {
+		n.Properties[NotificationPropertyOpniUuid] = util.HashStrings([]string{n.Title, n.Body})
+	}
+	return nil
+}
+
+func (l *ListNotificationRequest) Validate() error {
```
kralicky, line 522: Validate() shouldn't have side effects, setting defaults should happen elsewhere

pkg/test/environment.go
```diff
@@ -805,7 +806,7 @@ func (e *Environment) StartEmbeddedAlertManager(
 	if opniPort != nil {
 		defaultArgs = append(defaultArgs, fmt.Sprintf("--opni.listen-address=:%d", *opniPort))
 	}
-	_, cancelFunc := context.WithCancel(ctx)
+	_, caF = context.WithCancel(ctx)
```
kralicky, line 809: I don't think this will do anything if you discard the returned context

plugins/alerting/pkg/alerting/ops/ops.go
```diff
@@ -163,3 +166,22 @@ func (a *AlertingOpsNode) GetAvailableEndpoint(ctx context.Context, options *sha
...
+	status, err := a.GetClusterConfiguration(ctx, &emptypb.Empty{})
+	if err != nil {
+		return "", err
+	}
+	if status.NumReplicas == 1 { // exactly one that is the controller
+		availableEndpoint = options.GetControllerEndpoint()
+	} else {
+		availableEndpoint = options.GetWorkerEndpoint()
+	}
+	addr, _ := strings.CutPrefix(availableEndpoint, "http://")
```
kralicky, line 181:
```suggestion
	_, addr, _ := strings.Cut(availableEndpoint, "://")
```

plugins/alerting/pkg/alerting/ops/reconcile.go
```diff
@@ -3,17 +3,26 @@ package ops
...
+	"github.com/rancher/opni/pkg/util"
 	"github.com/rancher/opni/plugins/alerting/pkg/apis/alertops"
 	"google.golang.org/grpc/codes"
 	status "google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
 	"gopkg.in/yaml.v2"
 )
 
+func cacheEndpointToWebhook(endpoint string) string {
+	return fmt.Sprintf("http://%s", path.Join(endpoint, shared.AlertingDefaultHookName))
```
kralicky, line 23: this seems error-prone. also why http?

plugins/slo/pkg/slo/slo_test.go
```diff
@@ -13,7 +12,10 @@ import (
 	prommodel "github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/rulefmt"
 	"github.com/prometheus/prometheus/promql/parser"
+	v1 "github.com/rancher/opni/pkg/apis/capability/v1"
```
kralicky, line 15:
```suggestion
	capabilityv1 "github.com/rancher/opni/pkg/apis/capability/v1"
```

plugins/slo/pkg/slo/slo_test.go
```diff
@@ -73,7 +75,6 @@ var _ = Describe("Converting SLO information to Cortex rules", Ordered, Label(te
 			},
```
kralicky, line 75: This test suite should be labeled `integration`, also need to use the string literal labels instead of the constants. `Label("integration", "slow")`

plugins/slo/pkg/slo/slo_test.go
```diff
@@ -586,13 +606,14 @@ var _ = Describe("Converting SLO information to Cortex rules", Ordered, Label(te
 		Specify("The individual parts of the raw SLI queries should return data from cortex", func() {
```
kralicky, line 606: there is a todo comment a few lines above here that we should address

plugins/slo/pkg/slo/slo_test.go
```diff
@@ -586,13 +606,14 @@ var _ = Describe("Converting SLO information to Cortex rules", Ordered, Label(te
...
-			time.Sleep(time.Second)
 			rawGood, err := sloObj.RawGoodEventsQuery("5m")
 			Expect(err).NotTo(HaveOccurred())
 			rawTotal, err := sloObj.RawTotalEventsQuery("5m")
 			Expect(err).NotTo(HaveOccurred())
-
-			respGood, err := adminClient.Query(ctx, &cortexadmin.QueryRequest{
+			Eventually(func() error {
+				return nil
+			}, time.Second*90, time.Second*30).Should(BeNil())
```
kralicky, line 615: this seems extremely long

test/integration/alerting/alerting_test.go
```diff
@@ -518,24 +539,49 @@ func BuildAlertingClusterIntegrationTests(
...
+								Expect(len(messages.Items)).To(BeNumerically(">", 0))
+
+							} else {
+								if len(item.Windows) != 0 {
+									return fmt.Errorf("conditions that have not fired should not show up on timeline, but do")
+								}
+							}
+						}
+						return nil
+					}, time.Minute*2, time.Second*15)
```
kralicky, line 584: this also seems extremely long

kralicky (changes requested):

plugins/slo/pkg/slo/slo_test.go
alexandreLamarre: it has to do with recording rule evaluation times, these tests are not run in CI

plugins/alerting/pkg/alerting/ops/reconcile.go
alexandreLamarre: this is stop-gap solution to vanilla AlertManager only exposing cluster-peer addresses in its memberlist api, https://github.com/rancher/opni/issues/1150 will address these error prone definitions

plugins/slo/pkg/slo/slo_test.go
alexandreLamarre: prefer addressing it as a separate PR

pkg/apis/alerting/v1/extensions.go
alexandreLamarre: yep

plugins/alerting/pkg/alerting/ops/ops.go
alexandreLamarre: when run in kubernetes this cuts the service name out of the endpoint, I'll keep it as `http://` since it is more prudent

kralicky:
Test length issues can be dealt with as part of https://github.com/rancher/opni/issues/1171

go.mod
```diff
@@ -103,7 +105,7 @@ require (
 	github.com/qmuntal/stateless v1.6.2
 	github.com/rancher/charts-build-scripts v0.0.0-00010101000000-000000000000
 	github.com/rancher/kubernetes-provider-detector v0.1.5
-	github.com/samber/lo v1.37.0
+	github.com/samber/lo v1.36.1-0.20230220015530-adcf8499eb6a
```
kralicky, line 108: Should probably fix this

pkg/alerting/drivers/cortex/cortex.go
```diff
@@ -54,28 +49,34 @@ func ConstructFiltersFromMap(in map[string]string) string {
...
 		return nil, err
 	}
-	recordingRuleFmt := &rulefmt.Rule{
-		Record:      ConstructRecordingRuleName(info.GoldenSignal(), info.AlertType()),
-		Expr:        alertingRule.Expr,
+	recordingRuleFmt := &rulefmt.RuleNode{
+		Record: yaml.Node{
+			Kind:  yaml.ScalarNode,
+			Value: ConstructRecordingRuleName(info.GoldenSignal(), info.AlertType()),
+		}, //ConstructRecordingRuleName(info.GoldenSignal(), info.AlertType()),
```
kralicky, line 70: remove commented out code

kralicky (changes requested):

pkg/alerting/drivers/cortex/alert_matrix.go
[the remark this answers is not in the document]
alexandreLamarre: they're in `pkg/alerting/drivers/cortex/cortex_test.go`

kralicky (approved):

PR #1104, opened 2023-02-21
Implement http caching transport

### LINKED ISSUE #1021: HTTP transport caching

alexandreLamarre:
Closes #1021

### DIFF
--- a/pkg/caching/caching.go
+++ b/pkg/caching/caching.go
@@ -1,0 +1,56 @@
+package caching
+
+import (
+	"time"
+
+	"github.com/karlseguin/ccache"
+	"github.com/rancher/opni/pkg/storage"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+type InMemoryHttpTtlCache struct {
+	*ccache.Cache
+
+	maxAge time.Duration
+}
+
+func (i InMemoryHttpTtlCache) MaxAge() time.Duration {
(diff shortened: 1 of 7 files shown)


pkg/caching/caching.go
```diff
@@ -0,0 +1,56 @@
...
+
+	return item.Value().([]byte), true
+}
+
+func (i InMemoryHttpTtlCache) Set(key string, req []byte) {
+	i.Cache.Set(key, req, i.maxAge)
+}
+
+func (i InMemoryHttpTtlCache) Delete(key string) {
+	i.Cache.Delete(key)
```
dbason, line 36: Don't need to override this method as this will already be the behaviour of `i.Delete`

dbason (changes requested):

pkg/caching/caching.go
alexandreLamarre: It's a little weird but the underlying implementation of the eviction cache has 
```go
// Remove the item from the cache, return true if the item was present, false otherwise.
func (c *Cache) Delete(key string) bool {
```
which doesn't match the interface for the `httpcache` library, which expects nothing

kralicky (approved):

dbason (approved):

PR #1105, opened 2023-02-21
OEP: Allow user to specify anomalous keywords

AmartC:


codyrancher:
A couple notes to talk about

1.  The language "filter out" is being used in multiple places. Typically that means omit the matching items which I think is the opposite of the intent here.

1. We already sort of do this in opensearch plugin. Are we wanting to remove that in favor of this?

1. I wasn't a fan of doing this when we did it for the plugin. It's really odd to me for the user to tell us what's anomalous .

    1. The ML in theory should be doing this for the user.
    1. If the user is specifying it then it's not actually an anomaly. Instead, Ii's just keyword matching. Why are we conflating the two concepts?

 1. If we insist on doing this then I think I agree that this should be another page just to keep things simple.
 1. There was mention of workloads in here. Aren't we doing deployments instead?

PR #1107, opened 2023-02-22
fix issue with nil namespace in discovery request

joshmeranda:
Running `opni import discover` without a namespace filter causes `nil pointer dereference`. This PR resolves that oversight.

### DIFF
--- a/plugins/metrics/pkg/agent/node.go
+++ b/plugins/metrics/pkg/agent/node.go
@@ -224,7 +224,12 @@ func (m *MetricsNode) Discover(ctx context.Context, request *remoteread.Discover
 		}, nil
 	}
 
-	entries, err := m.nodeDriver.DiscoverPrometheuses(ctx, *request.Namespace)
+	var namespace string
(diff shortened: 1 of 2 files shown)


plugins/metrics/pkg/agent/node.go
```diff
@@ -224,7 +224,12 @@ func (m *MetricsNode) Discover(ctx context.Context, request *remoteread.Discover
 		}, nil
 	}
 
-	entries, err := m.nodeDriver.DiscoverPrometheuses(ctx, *request.Namespace)
+	var namespace string
+	if request.Namespace != nil {
```
alexandreLamarre, line 228: nit : consider using https://pkg.go.dev/github.com/samber/lo#readme-fromptror

joshmeranda: Nice, I couldn't find a util method for this yesterday, thanks!

alexandreLamarre (approved):

PR #1131, opened 2023-03-03
Etl plugin

joshmeranda:
Move ETL tool implementation to new plugin separate from metrics to avoid issues with existing PrometheusOperators.


joshmeranda:
closing in favor of using https://github.com/rancher/opni/pull/1210 in the future to support cleaner enable / disable of `kube-stack-prometheus` at install time

PR #1149, opened 2023-03-07
Switch to use OTEL for logging agent

### LINKED ISSUE #1127: Deploy new logging with Agent
When logging capability is installed the new logging infra should be deployed.  Agent should migrate to the new infra when it is upgraded.

Acceptance Criteria

- [ ] Agent sync creates new custom resources
- [ ] Agent sync deletes old custom resources if they exist

dbason:
Fixes #1127


apis/core/v1beta1/collector_meta.go
```diff
@@ -0,0 +1,10 @@
+package v1beta1
+
+func (c *CollectorSpec) IsEmpty() bool {
+	empty := true
+	if c.LoggingConfig != nil {
+		empty = false
+	}
+
+	return empty
```
alexandreLamarre, line 9: can probably simplify this to `return c.LoggingConfig != nil`

dbason: So currently yes, but this function is setting the foundation for future work.  In the near future we will add metrics configuration, and it should only return true if there are neither logging nor metrics config.  So in this case we would have an additional if block e.g.
```go
if c.MetricsConfig != nil {
    empty = false
}
```

alexandreLamarre: its definitely a stylistic nitpick but : `c.Metrics == nil && c.LoggingConfig == nil && ...`  works just as well

Issue #1162, opened 2023-03-10
OTEL Metrics Collection : Kube metrics


alexandreLamarre:
Related to [OEP](https://github.com/rancher/opni/blob/main/enhancements/monitoring/20230117-move-to-otel-collector.md)

alexandreLamarre:
In progress, rolling over to sprint 4

PR #1176, opened 2023-03-13
Collect kubernetes + workload metrics with OTEL

alexandreLamarre:
Closes https://github.com/rancher/opni/issues/1162


magefile.go
```diff
@@ -107,7 +110,40 @@ func Build() error {
 }
 
 func Generate() {
-	mg.SerialDeps(Protobuf, mockgen.Mockgen, ControllerGen)
+	mg.SerialDeps(Protobuf, mockgen.Mockgen, ControllerGen, GenerateOTEL)
+}
+
+func GenerateOTEL() error {
```
dbason, line 116: Can probably remove this seeing as we have it in a separate repo

pkg/otel/templates/base.tmpl
```diff
@@ -0,0 +1,67 @@
+{{- define "aggregator-config" -}}
```
dbason, line 1: Do we actually use this template anywhere?

pkg/otel/templates/base.tmpl
```diff
@@ -0,0 +1,67 @@
...
+  {{template "metrics-self-telemetry" . }}
+  pipelines:
+  {{template "metrics-remotewrite-pipeline" . }}
+{{- end -}}
+
+{{- define "node-receivers" -}}
+{{- template "metrics-node-receivers" . -}}
+{{- end -}}
+
+{{define "node-config" }}
```
dbason, line 43: Likewise are we actually using this template anywhere?

dbason (changes requested):

pkg/otel/templates/base.tmpl
alexandreLamarre: config validation tests for now

pkg/otel/templates/base.tmpl
alexandreLamarre: also config validation tests for now

magefile.go
alexandreLamarre: We can probably disassociate it from the generate Target and rename it to something like fetchOTELDistro, but still convenient seeing as we do require the binary for running otel in the test environment.

Or, I could add something like `goreleaser` to the sandbox repo for releases and add it those binaries to our testbin target

pkg/otel/templates/base.tmpl
alexandreLamarre: will move these to pkg/test/testdata to be more idiomatic

magefile.go
alexandreLamarre: https://github.com/rancher-sandbox/opni-otel-collector/pull/3

dbason: Yeah I think fetching the binaries is the better way to go.  An alternative to attaching the binary would be using `go get` to fetch and build the binary

pkg/otel/templates/base.tmpl
dbason: These feel a little clunky to me because they're completely disconnected from the actual code.  E.g it's possible to modify the actual templates the code is deploying without changing these and be in a situation where the validation is passing, but the templates themselves won't (unless I'm missing something).  The instinct is right, but I think this is not the right mechanism.

alexandreLamarre: yeah, I couldn't come up with anything better atm, open to any suggestions

alexandreLamarre: we could move all the templates from pkg/resources/collector into the fs.Embed and have the controller tests build the config maps and drive the config validation there?

dbason: To be honest I would just remove them until we come up with something better.

dbason (approved):

pkg/otel/templates/base.tmpl
alexandreLamarre: Started https://github.com/rancher/opni/discussions/1245 to form ideas over a long term solution

dbason (approved):

pkg/test/testdata/prometheus/otel-config.yaml
```diff
@@ -0,0 +1,44 @@
...
+    limit_mib: 250
+    spike_limit_mib: 50
+    check_interval: 1s
+  resourcedetection/system:
+    detectors: ["system"]
+    system:
+      hostname_sources: ["lookup"]
+exporters:
+  prometheusremotewrite:
+    endpoint: "http://localhost:{{ .OpniAgentPort }}/api/agent/push"
```
kralicky, line 30: heads up this changes in https://github.com/rancher/opni/pull/1271, will need to be
```suggestion
    endpoint: "http://{{ .OpniAgentAddress }}/api/agent/push"
```

alexandreLamarre:
requires #1271

magefile.go
alexandreLamarre: I ended up adding gorelease to the sandbox, and using a testbin target, since it falls in line with other things in the test env

apis/monitoring/v1beta1/collector_config_types.go
```diff
@@ -0,0 +1,52 @@
...
+	NamespaceSelector      []string
+	ServiceMonitorSelector []string
+	PodMonitorSelector     []string
+	ProbeSelector          []string
+}
+
+type CollectorConfigSpec struct {
+	PrometheusDiscovery PrometheusDiscovery `json:"prometheusDiscovery,omitempty"`
+	RemoteWriteEndpoint string              `json:"remoteWriteEndpoint,omitempty"`
+	OtelSpec            node.OTELSpec       `json:"otelSpec,omitempty"`
```
kralicky, line 18: copy the OTELSpec struct (or an equivalent refactor), such that we don't have to import plugins/metrics from here

pkg/otel/types.go
```diff
@@ -1,11 +1,87 @@
 package otel
 
-import "fmt"
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/rancher/opni/plugins/metrics/pkg/apis/node"
```
kralicky, line 7: Can't import plugins/metrics from here

plugins/metrics/pkg/agent/plugin.go
```diff
@@ -43,6 +43,10 @@ func NewPlugin(ctx context.Context) *Plugin {
 		return drivers.NewExternalPromOperatorDriver(lg)
 	})
 
+	drivers.RegisterNodeDriverBuilder("otel-node-driver", func() (drivers.MetricsNodeDriver, error) {
```
kralicky, line 46: suggest renaming to something like "opni-manager-otel", what do you think?

kralicky (changes requested):

plugins/metrics/pkg/agent/plugin.go
alexandreLamarre: no objection :+1:

kralicky (approved):

PR #1197, opened 2023-03-15
Fix jetstream storage config reconciling

### LINKED ISSUE #1195: Panic instead of error log when missing jetstream config options
When using the jetstream storage backend but the jetstream options are not configured, it should return a human-readable error, but instead panics later on when trying to use the nil config field.

kralicky:
Fixes #1195


alexandreLamarre (dismissed):

pkg/resources/gateway/configmap.go
```diff
@@ -90,6 +111,18 @@ func (r *Reconciler) configMap() (resources.Resource, string, error) {
...
+		if err := r.client.Get(r.ctx, types.NamespacedName{
+			Namespace: r.gw.Namespace,
+			Name:      r.gw.Spec.NatsRef.Name,
+		}, nats); err != nil {
+			return nil, "", err
+		}
+		gatewayConf.Spec.Storage.JetStream = &cfgv1beta1.JetStreamStorageSpec{
+			Endpoint:     fmt.Sprintf("nats://%s-nats-client.%s.svc:%d", nats.Name, nats.Namespace, natsutil.DefaultClientPort),
+			NkeySeedPath: fmt.Sprintf("%s/seed", natsutil.NkeyDir),
+		}
```
dbason, line 125: I feel like we could probably reuse https://github.com/rancher/opni/blob/main/pkg/util/nats/nats.go#L28 here

dbason (changes requested):

dbason (approved):

Issue #1126, opened 2023-03-02
[EPIC] Human Readable Logging Cluster Names


joshmeranda:
## Summary: 
Currently, when users select an agent cluster for log analytics, they select from a list of cluster IDs. This makes it difficult to select the right agent among many. This proposal aims to add human-readable agent cluster names, to help differentiate the cluster names for the user.

In the Opni UI, we will adda a new "Clusters" tab which will be responsible for displaying the available clusters and their metadata to the user.

 ## Use case: 
When an Opni Gateway has many connected agents, it becomes cumbersome to search through a list of IDs to find the one which matches the target agent.

 ## Benefits: 
 By viewing a list of human-readable names, it will be immediately clear which cluster you are selecting. While the IDs do uniquely identify each agent, it's not immediately clear what responsibilities each agent has. Human-readable names will make agent responsibility clearer, as well as being easier to remember.

## Level of effort: 
This addition should take 1.5 weeks for implementation and testing. The aim is to have this up for the March 15th release.

## Issues
* https://github.com/rancher/opni/issues/1057
* https://github.com/rancher/opni/issues/1115
* https://github.com/rancher/opni/issues/1116
* https://github.com/rancher/opni/issues/1117

sanjay920:
On track after sprint 3

Issue #1116, opened 2023-02-27
On cluster rename, the cluster metadata index should be updated with the new name


joshmeranda:

joshmeranda:
Being tracked in: https://github.com/rancher/opni/issues/1126

Issue #1115, opened 2023-02-27
When a cluster is added, ad a new entry to the cluster metadata index


joshmeranda:

joshmeranda:
Being tracked in: https://github.com/rancher/opni/issues/1126

PR #1218, opened 2023-03-20
Human Readable Logging Names

joshmeranda:
https://github.com/rancher/opni/issues/1115
https://github.com/rancher/opni/issues/1116
https://github.com/rancher/opni/issues/1126

https://github.com/rancher/opni/pull/969


plugins/logging/pkg/gateway/system.go
```diff
@@ -72,3 +78,93 @@ func (p *Plugin) UseKeyValueStore(client system.KeyValueStoreClient) {
 	p.uninstallController.Set(ctrl)
 	<-p.ctx.Done()
 }
+
+func (p *Plugin) watchGlobalCluster(client managementv1.ManagementClient) {
+	clusterClient, err := client.WatchClusters(p.ctx, &managementv1.WatchClustersRequest{})
```
alexandreLamarre, line 83: consider listing clusters before the for loop, to check they match the expected state on startup

plugins/logging/pkg/gateway/system.go
```diff
@@ -72,3 +78,93 @@ func (p *Plugin) UseKeyValueStore(client system.KeyValueStoreClient) {
...
+	if err != nil {
+		p.logger.Errorf("failed to watch clusters, exiting")
+		os.Exit(1)
+	}
+
+	p.logger.Infof("watching cluster events")
+
+	for {
```
alexandreLamarre, line 91: @kralicky  @dbason we may want "watchers" to persist their last received(/successfuly handled?) state (in the plugin KV), so that if consumers of the events are disconnected or not running, they can quickly check state differences

pkg/gateway/watcher.go
```diff
@@ -0,0 +1,26 @@
+package gateway
+
+type InternalConditionWatcher interface {
+	WatchEvents()
```
alexandreLamarre, line 4: may want to have this interface implement contracts for something like `RegisterEvent(eventType, funcs ...)` & `HandleEvent(eventType)`?

joshmeranda: @alexandreLamarre Added, let me know if I missed anything.

plugins/logging/pkg/gateway/system.go
dbason: I think we may want to to implement an event streaming light type pattern.  We could keep a history of events and these could be replayed on demand.  There are two options here; we could use a NATSs stream, or we could just handle this inside opni initially.

The other thing we could potentially do is have a sync initial state function which could be used on startup.  This would reset the event history and just pass out the starting state.

pkg/gateway/watcher.go
alexandreLamarre: LGTM, should probably add unit tests for these

plugins/logging/pkg/gateway/system.go
alexandreLamarre: Cool, we can probably omit a replay feature until we have more complex events that need something like that, my [other comment](https://github.com/rancher/opni/pull/1218#discussion_r1143485824) suggested something along the lines of an initial sync.

joshmeranda: Makes sense, to get this feature in for the next release I might go for the initial sync between opensearch index and the management cluster approach and we can write up an OEP if we want to start adding event replays for plugins down the line.

plugins/logging/pkg/backend/metadata.go
```diff
@@ -0,0 +1,227 @@
...
+		b.Logger.With(zap.Error(err)).Errorf("failed to update cluster in metadata index")
+		return nil
+	}
+
+	if resp.IsError() {
+		errMsg, err := io.ReadAll(resp.Body)
+
+		if err != nil {
+			b.Logger.With(zap.Error(err)).Errorf("failed to read response body")
+			return nil
```
alexandreLamarre, line 189: potential resource leak in this path : resp.Body is never closed
for the functions in this file, consider using the pattern:
```go
resp, err := // ...
if err != nil {
  // ...
}
defer resp.Body.Close
```

joshmeranda: I call `resp.Body.Close()` down below, but deferring is definitely more idiomatic

alexandreLamarre: there is a `return nil` in that particular code path that results in the resp.Body.Close below not being closed

alexandreLamarre (dismissed):

pkg/gateway/watcher.go
```diff
@@ -0,0 +1,64 @@
...
+
+func NewConditionWatcher(cl ...func()) ConditionWatcher {
+	return &internalWatcher{
+		closures: cl,
+	}
+}
+
+func (w *internalWatcher) WatchEvents() {
+	for _, cl := range w.closures {
+		f := cl
```
dbason, line 25: Nit: why are we creating a new variable here?

pkg/management/clusters.go
```diff
@@ -91,21 +91,26 @@ func (m *Server) WatchClusters(
 
 	for event := range eventC {
 		var c *corev1.Cluster
+		var o *corev1.Cluster
```
dbason, line 94: Nit: cleaner variable declaration here might be `var c, o *corev1.Cluster`

plugins/logging/pkg/backend/metadata.go
```diff
@@ -0,0 +1,269 @@
...
+	"golang.org/x/sync/errgroup"
+	"io"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"net/http"
+	"os"
+	"time"
+)
+
+func (b *LoggingBackend) waitForOpensearchClient(ctx context.Context) error {
+	var retErr error
```
dbason, line 25: is there any reason not to use `b.WaitForInit()` here?  If we want a method that handles a context closing I think it would be better off adding it in there.

plugins/logging/pkg/backend/metadata.go
```diff
@@ -0,0 +1,269 @@
...
+			close(stopChan)
+		} else {
+			b.Logger.Errorf("opensearch client not set, waiting")
+		}
+	}, time.Second, stopChan)
+
+	return retErr
+}
+
+func (b *LoggingBackend) reconcileClusterDoc(ctx context.Context, doc *types.ClusterMetadataDoc, cluster *opnicorev1.Cluster) error {
```
dbason, line 45: We should move the logic of this into the opensearch data manager itself.  We could do the wait for client inside of that as well.

plugins/logging/pkg/backend/metadata.go
```diff
@@ -0,0 +1,269 @@
...
+		})
+
+		if !hasLogging {
+			continue
+		}
+
+		cluster := cluster
+
+		eg.Go(func() error {
+			resp, err := b.OpensearchManager.Indices.GetDocument(egCtx, multiclusterrolebinding.ClusterMetadataIndexName, cluster.Id)
```
dbason, line 96: Again we should put this logic inside the data manager

plugins/logging/pkg/backend/metadata.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		return fmt.Errorf("could not get data for cluster '%s': %w", ref.Id, err)
+	}
+
+	clusterName, found := cluster.Metadata.Labels[opnicorev1.NameLabel]
+	if !found {
+		clusterName = cluster.Id
+	}
+
+	resp, err := b.OpensearchManager.Indices.AddDocument(ctx, multiclusterrolebinding.ClusterMetadataIndexName, cluster.Id, opensearchutil.NewJSONReader(map[string]string{
```
dbason, line 163: Again I think this logic is better inside the data manager

plugins/logging/pkg/backend/metadata.go
```diff
@@ -0,0 +1,269 @@
...
+	).Debug("cluster was renamed")
+
+	if b.OpensearchManager.Client == nil {
+		b.Logger.With(
+			"cluster", clusterId,
+		).Warnf("plugin has nil opensearch client, doing nothing")
+		return nil
+	}
+
+	resp, err := b.OpensearchManager.Indices.UpdateDocument(ctx, multiclusterrolebinding.ClusterMetadataIndexName, clusterId, opensearchutil.NewJSONReader(
```
dbason, line 212: Again put this inside the data manager

dbason (changes requested):

plugins/logging/pkg/backend/metadata.go
joshmeranda: Not really. The only reason we might want to avoid it here is if we have to add more init code that relies on the opensearch client and need to move the call to `waitForOpensearchClient` outside of the goroutine [here](https://github.com/rancher/opni/blob/human-logging/plugins/logging/pkg/backend/logging.go#L63-L75) which would block `LoggingBackend.Initialize(...)` indefinitely. But this is super unlikely. I'll add `b.WaitForInit()`

pkg/gateway/watcher.go
joshmeranda: If we don't create `f` we risk running the wrong function in the following closure since the goroutine would capture the current value of `cl` when the goroutine starts rather than at the time the goroutine was declared. So we need to copy the func to ensure that we are calling the right function in the goroutine.

https://go.dev/play/p/Q_OTVaYclXB

alexandreLamarre (dismissed):

pkg/gateway/watcher.go
kralicky: The form `cl := cl` is more common in our codebase, that's what I would recommend

pkg/gateway/watcher.go
kralicky, line 1: This might be better suited in either apis/management/v1 or pkg/management, Imports directly into pkg/gateway should be kept to a minimum

joshmeranda: works for me, moved to `pkg/management`

pkg/gateway/watcher.go
joshmeranda: @kralicky renamed :+1:

pkg/management/watcher_test.go
```diff
@@ -0,0 +1,43 @@
+package management
+
+import (
+	"context"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	managementv1 "github.com/rancher/opni/pkg/apis/management/v1"
+)
+
+var _ = Describe("ManagementWatcherHooks", Label("unit"), func() {
```
dbason, line 10: Needs to have the `Ordered` argument to run `BeforeAll` containers.

dbason (changes requested):

pkg/apis/management/v1/management.proto
```diff
@@ -235,6 +235,7 @@ enum WatchEventType {
 message WatchEvent {
   core.Cluster cluster = 1;
   WatchEventType type = 2;
+  core.Cluster previousCluster = 3;
```
kralicky, line 238: Recommend naming this just `previous` for consistency with other message types

dbason (approved):

Issue #1227, opened 2023-03-21
Add Grafana support in the test environment


kralicky:
The test environment should be able to generate a Grafana configuration and (optionally) `docker run` a Grafana container that connects to the running test env.

Issue #1219, opened 2023-03-20
Agent stream metrics instrumentation


kralicky:
We should add a few new prometheus metrics exposed by the gateway that can be used to monitor networking info (throughput, latency, etc) for agent grpc streams.

Issue #1214, opened 2023-03-16
Remote Write API improvements


kralicky:
In order to efficiently intercept remote write payloads, the RemoteWrite metrics API needs to be changed to utilize native prometheus/cortex WriteRequest protobuf types.

Issue #1157, opened 2023-03-08
Per-tenant Cluster Info Metrics


sanjay920:
OEP: https://github.com/rancher/opni/blob/main/enhancements/monitoring/20230131-tenant-impersonation.md
Using the new local agent identification feature, modify the `opni_cluster_info` metric to use tenant impersonation as described in https://github.com/rancher/opni/pull/1013

kralicky:
This can be started on now that https://github.com/rancher/opni/pull/1060 and https://github.com/rancher/opni/pull/1132 are completed.

kralicky:
In order to efficiently intercept remote write payloads, I'm going to update the metrics plugin to use native prometheus write request protobuf types. This should also be mostly reusable for otel metrics format in the future.

kralicky:
For easier testing of the new metrics, I am adding Grafana support to the test env. Tracked in https://github.com/rancher/opni/issues/1227

PR #1229, opened 2023-03-21
Tenant Impersonation + Per-tenant cluster info metrics

kralicky:
Closes https://github.com/rancher/opni/issues/1157
Closes https://github.com/rancher/opni/issues/1227
Closes https://github.com/rancher/opni/issues/1219
Closes https://github.com/rancher/opni/issues/1214
Closes https://github.com/rancher/opni/issues/1253


pkg/test/environment.go
```diff
@@ -1898,13 +1917,96 @@ func (e *Environment) JetStreamConfig() *v1beta1.JetStreamStorageSpec {
...
+		panic(err)
+	}
+}
+
+func (e *Environment) StartGrafana() {
+	baseDir := path.Join(e.tempDir, "grafana")
+
+	args := []string{
+		"run",
+		"-q",
```
alexandreLamarre, line 1963: Would be nice to set some runtime resource limits for this

pkg/test/testdata/cortex/config.yaml
```diff
@@ -99,3 +99,9 @@ ruler_storage:
   backend: filesystem
   filesystem:
     dir: "{{ .StorageDir }}/rules"
+
+limits:
+  metric_relabel_configs:
```
alexandreLamarre, line 104: is it possible to marshal the configuration used in cortex/config.go here for consistency?
```
MetricRelabelConfigs: []*relabel.Config{
				metrics.OpniInternalLabelFilter(),
			},
```

plugins/metrics/pkg/agent/http.go
```diff
@@ -57,79 +57,103 @@ func (s *HttpServer) SetRemoteWriteClient(client clients.Locker[remotewrite.Remo
...
 	})
-
 	if !ok {
-		c.Status(http.StatusServiceUnavailable)
+		return nil, status.Errorf(codes.Unavailable, "gateway not connected")
 	}
+	return
 }
+
+// func (s *HttpServer) handleMetricPushRequest(c *gin.Context) {
```
joshmeranda, line 87: Do we still need this commented code or can it be dropped?

alexandreLamarre, line 87: commented out block can probably be removed

plugins/metrics/pkg/apis/remotewrite/remotewrite.proto
```diff
@@ -4,18 +4,19 @@ option go_package = "github.com/rancher/opni/plugins/metrics/pkg/apis/remotewrit
...
 
 package remotewrite;
 
-// The RemoteWrite service is used by metrics nodes to post Prometheus metrics 
+// The RemoteWrite service is used by metrics nodes to post Prometheus metrics
 // and rules to Cortex.
 service RemoteWrite {
-  rpc Push(Payload) returns (google.protobuf.Empty);
+  rpc Push(cortexpb.WriteRequest) returns (cortexpb.WriteResponse);
+  // rpc SyncRules(rules.RuleDesc) returns (google.protobuf.Empty);
```
joshmeranda, line 15: more commented code that could be pulled out

plugins/metrics/pkg/apis/remotewrite/compat.go
```diff
@@ -0,0 +1,68 @@
...
+func AsPrometheusRemoteWriteClient(rwc RemoteWriteClient) PrometheusRemoteWriteClient {
+	switch rwc.(type) {
+	case *remoteWriteClient:
+		return prometheusRemoteWriteClient{cc: rwc.(*remoteWriteClient).cc}
+	default:
+		return slowPrometheusRemoteWriteClient{rwc: rwc}
+	}
+}
+
+func labelProtosToLabels(labelPairs []prompb.Label) labels.Labels {
```
alexandreLamarre, line 58: might be useful to have these prometheus struct conversion in pkg/metrics/util or pkg/util?  I could see alerting & other metric enhancements wanting to use this down the line.

If you agree we should probably also move https://github.com/rancher/opni/blob/cc837d7280678ffcdba3b122e56173f07c7c5622/plugins/metrics/pkg/agent/runner.go#L28 to the same package

plugins/metrics/pkg/cortex/intercept.go
```diff
@@ -0,0 +1,192 @@
...
+	defaultId, err := user.ExtractOrgID(outgoingCtx)
+	if err != nil {
+		panic(err)
+	}
+	var buckets = map[string][]cortexpb.PreallocTimeseries{
+		defaultId: {},
+	}
+TIMESERIES:
+	for _, ts := range req.Timeseries {
+		nameLabelIndex := -1
```
alexandreLamarre, line 85: can simplify this block using [lo.FindIdexOf](https://pkg.go.dev/github.com/samber/lo#readme-findindexof)

plugins/metrics/pkg/cortex/intercept.go
```diff
@@ -0,0 +1,192 @@
...
+			continue
+		}
+		for i, l := range ts.Labels {
+			if l.Name == t.conf.IdLabelName {
+				if t.conf.DropIdLabel {
+					ts.Labels = append(ts.Labels[:i], ts.Labels[i+1:]...)
+				}
+				buckets[l.Value] = append(buckets[l.Value], ts)
+				t.mdTracker.track(ts.Labels[nameLabelIndex].Value, l.Value)
+				continue TIMESERIES
```
alexandreLamarre, line 103: this seems potentially expensive in the worst case (high volume of metrics, high number of labels, many drops)

alexandreLamarre: that scenario may never realistically happen, but maybe it makes sense to eventually add metrics for the federating interceptor itself, kinda like prometheus's scrape metrics

plugins/metrics/pkg/apis/remotewrite/compat.go
alexandreLamarre: Alerting will probably end having utility method for `func convert(labels []*prompb.LabelMatcher)[]labels.LabelMatcher` for its V2 apis

pkg/test/testdata/cortex/config.yaml
kralicky: Probably not, the actual production config is too different than the test one.

plugins/metrics/pkg/apis/remotewrite/compat.go
kralicky: Yeah i can move some of these around

plugins/metrics/pkg/cortex/intercept.go
kralicky: I can try to optimize this a bit.

kralicky: Working on optimizing this significantly, since it is in the hot path -- making good progress.

plugins/metrics/pkg/cortex/intercept.go
```diff
@@ -0,0 +1,357 @@
...
+	defaultId, err := user.ExtractOrgID(outgoingCtx)
+	if err != nil {
+		panic(err)
+	}
+
+	startTime := time.Now()
+
+	// save the original slice so we can return it to the pool, along with all
+	// the original timeseries entries
+	allTimeseries := req.Timeseries
```
alexandreLamarre, line 91: as far as I can tell, the `:=` operator does not perform slice copies, so either the behavior and comment are mismatched or the comment is confusing here, maybe some additional clarification is needed?

plugins/metrics/pkg/cortex/intercept.go
```diff
@@ -0,0 +1,357 @@
...
+		toReturn := status.Convert(errs[0])
+		for _, err := range errs[1:] {
+			if statErr := status.Convert(err); statErr.Code() > toReturn.Code() {
+				toReturn = statErr
+			}
+		}
+		return nil, toReturn.Err()
+	}
+
+	req.Timeseries = allTimeseries
```
alexandreLamarre, line 161: see above comment, might be an an ineffective assign?

plugins/metrics/pkg/cortex/intercept.go
kralicky: This saves the slice itself (the pointer into the backing array + length), not the contents. We reuse the request message and change the slice bounds in each call to the callback function, but in order for the entire contents of the backing array to be returned to the pool correctly, we have to reset the slice to the original one before exiting.

plugins/metrics/pkg/cortex/admin.go
```diff
@@ -290,6 +290,32 @@ func (p *CortexAdminServer) QueryRange(
 	}, nil
 }
 
+func (p *CortexAdminServer) GetMetricMetadata(ctx context.Context, req *cortexadmin.MetricMetadataRequest) (*cortexadmin.MetricMetadata, error) {
```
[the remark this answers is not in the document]
kralicky: Yes, it will just have to query the store every time.

plugins/metrics/pkg/cortex/intercept.go
alexandreLamarre: I would change this comment from `// save the original slice`  to something along the lines of that explanation

alexandreLamarre (dismissed):

dbason (approved):

dbason (approved):

alexandreLamarre (approved):

Issue #1147, opened 2023-03-06
Update /model/train endpoint to enable AI services before sending over workloads.


AmartC:
When the /model/train endpoint is called, before the model payload is sent, it will first enable the AI services before sending over the payload to Nats to the AI services.

AmartC:
Rolled over to sprint 4.

AmartC:
PR has been approved and is awaiting approval of PRs to the training controller service and inferencing service before issue can be closed.

PR #1252, opened 2023-03-28
Update AIOps plugin to enable AI services when training job is submitted

AmartC:
The AIOps plugin has been updated to enable AI services when a training job is submitted. This PR addresses https://github.com/rancher/opni/issues/1147

### DIFF
--- a/plugins/aiops/pkg/gateway/modeltraining.go
+++ b/plugins/aiops/pkg/gateway/modeltraining.go
@@ -7,8 +7,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/nats-io/nats.go"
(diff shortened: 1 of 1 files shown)


plugins/aiops/pkg/gateway/modeltraining.go
```diff
@@ -26,8 +28,14 @@ const (
 	modelStatusNatsSubject     = "model_status"
 )
 
+type ModelTrainingParameters struct {
+	UUid      string                         `json:"uuid,omitempty"`
```
dbason, line 32: Nit: this should be `UUID`

dbason (changes requested):

dbason:
I think you should squash this commit into the previous one.  We don't need a separate commit for it.

dbason (dismissed):

dbason (approved):

PR #1254, opened 2023-03-28
OEP: Cortex Config Options

kralicky:
https://github.com/rancher/opni/issues/1262


sanjay920 (changes requested):
@kralicky @codyrancher can you add UI details to this OEP?

dbason (commented):
As part of this did we want to move S3 configuration to a generic cluster wide setup?

dbason (approved):

sanjay920 (approved):

PR #1282, opened 2023-04-04
Implement per-node metrics capability config

kralicky:
Closes https://github.com/rancher/opni/issues/1289
Fixes https://github.com/rancher/opni/issues/1288
Fixes https://github.com/rancher/opni/issues/1291
Fixes https://github.com/rancher/opni/issues/1292


pkg/opni/commands/metrics.go
```diff
@@ -14,6 +14,7 @@ func BuildMetricsCmd() *cobra.Command {
 
 	cmd.AddCommand(BuildCortexAdminCmd())
 	cmd.AddCommand(BuildCortexOpsCmd())
+	cmd.AddCommand(BuildMetricsConfigCmd())
```
joshmeranda, line 17: would it make more sense to have capabilty configuration commands under the `opni capabilty` command? Feels out of scope

kralicky: Plugin-specific commands are grouped together so that they can be compiled out using build tags. The current structure keeps the core commands separated. I agree it's not the most obvious structure. Refactoring would be out of scope for this PR but we can definitely take a look at updating this

joshmeranda: Ok works for me

joshmeranda (dismissed):

test/plugins/metrics/node_config_test.go
```diff
@@ -0,0 +1,275 @@
...
+	var getConfig = func(agentId string) (*node.MetricsCapabilitySpec, bool, error) {
+		var trailer metadata.MD
+		spec, err := nodeClient.GetNodeConfiguration(context.Background(), &v1.Reference{Id: agentId}, grpc.Trailer(&trailer))
+		if err != nil {
+			return nil, false, err
+		}
+		var isDefault bool
+		if len(trailer["is-default-config"]) > 0 {
+			isDefault = trailer["is-default-config"][0] == "true"
+		}
```
alexandreLamarre, line 64: might be worth putting this in a package since this default check is also used in the CLI

pkg/apis/capability/v1/validation.go
```diff
@@ -15,3 +15,13 @@ func (req *InstallRequest) Validate() error {
 	}
 	return nil
 }
+
+func (req *SyncRequest) Validate() error {
+	if req.Cluster != nil && req.Cluster.Id != "" {
+		// empty string indicates "all" here
```
alexandreLamarre, line 21: should add this comment to the SyncRequest message definition, for IDE completion

alexandreLamarre (dismissed):

alexandreLamarre (dismissed):

Issue #1163, opened 2023-03-10
OTEL Metrics Collection : Batching & Label Processing


alexandreLamarre:
Related to [OEP](https://github.com/rancher/opni/blob/main/enhancements/monitoring/20230117-move-to-otel-collector.md)

alexandreLamarre:
Consider removing some host metrics for more streamlined processing
- processes
- filesystem
- disk merges

alexandreLamarre:
not started, rolling over to sprint 4

alexandreLamarre:
K8s resource attribute labels contain valuable data for downstream use, but for metrics we should drop container metadata and resource uid labels during processing on the node otel agents to reduce memory footprint

alexandreLamarre:
since the `action` part of attribute processing process doesn't allow for matching keys to regex, we may need to "externalize" the portions of `k8s.*.uuid` into our struct that executes the templates

```
k8sattributes:
    passthrough: false
    pod_association:
    - sources:
      - from: resource_attribute
        name: k8s.pod.ip
    - sources:
      - from: resource_attribute
        name: k8s.pod.uid
    - sources:
      - from: connection
```
to explicitly drop all relevant `_uid` labels on metrics.
```yaml
  attributes/k8smetrics:
    include:
      match_type: regexp
      metric_names : ["^k8s"]
      attributes:
      - key: k8s_pod_uid
#...
    actions:
#...
      - action: delete
        key: k8s_pod_uid
```

alexandreLamarre:
or we have a separate  k8sattribute processor for metrics, but i think having the same shared context information for all observability data will also be important for aiops services

alexandreLamarre:
also looking through `node_exporter` & `kube_state` metrics exporters documentation (exporters bundled with kube-prometheus-stack) to see which labels we can drop on metrics to further reduce load

alexandreLamarre:
specifically for dropping labels on metrics we can't use attribute processors / metrictransforms/ base OTTLP functions on metrics, because dropping all values for a particular label is an unsafe operation (will the metric still be valid, how are data points associated after the drop?). We must do so with aggregations, but the aforementioned processors can only do aggregations when they know the result label set - which in our generic case is not possible.

We will have to either create our own OTTLP functions for the transform processor or include this logic in a metrics OTLP forward on the gateway. So will circle back to this issue once we have the gateway OTLP implementation and draw comparisons between the two methods.

alexandreLamarre:
[...]

PR #1308, opened 2023-04-14
Process OTLP metrics 

alexandreLamarre:
Addresses parts of #1163


alexandreLamarre:
This PR is somewhat stale, and was intended to further reduce the footprint of metrics collection, but in reality this doesn't matter.

PR #1322, opened 2023-04-19
Metric service

tybalex:


kralicky:
Should we track `model.pth` with git lfs?

aiops/metric/requirements.txt
kralicky, line 1: After `pip install -r requirements.txt` I was still missing the following: 
- fastapi
- torch
- requests

aiops/metric/tests/test_main.py
kralicky, line 1: We should add some more tests here e.g. interacting with the metrics plugin apis using the test environment (I can help with this)

aiops/metric/metric_analysis/cortexadmin_pb.py
kralicky, line 1: There's another copy of the generated code in aiops/apis/opni_proto, should everything be generated there instead?

kralicky (changes requested):

aiops/metric/requirements.txt
tybalex: these are installed in the base image `rancher/opni-python-base:3.8-torch`

aiops/metric/tests/test_main.py
tybalex: yep, I will add a few more.

tybalex:
> Should we track `model.pth` with git lfs?

yes

aiops/metric/metric_analysis/cortexadmin_pb.py
tybalex: yes.  I am only use this one temporarily.

aiops/metric/tests/test_main.py
tybalex: @kralicky I do need your help here in terms of `interacting with the metrics plugin apis using the test environment`. Since most of the APIs here would have to interact with either gateway API or natsKV or both.

aiops/metric/metric_analysis/cortexadmin_pb.py
tybalex: @kralicky maybe it's better to generate this file here, so we can always keep this piece of code update-to-date? `Opni_proto` is maintained manually so it requires extra efforts to update it with changes in gateway.

aiops/metric/tests/test_main.py
tybalex: I added unit-test for the model and metric_filter methods. for tests to interact with metric plugins we can cover in a different place

kralicky: unit tests look good, I'll help you set up the integration tests

aiops/metric/metric_analysis/cortexadmin_pb.py
tybalex: removed these 2 files -- cortexadmin_pb and status_pb. magefile will generate them into this dir automatically.

tybalex:
Regard to the model file, @AmartC  will move it to Opni's internal S3 bucket.

Issue #1153, opened 2023-03-07
AlertOps v2


alexandreLamarre:
- [ ] Deploy cortex alertmanager for alerting via opni binary
- [ ] V2 api of CortexOps that whose cluster configuration is determined by:
  - [ ] Advanced memberlist configuration
  - [ ] Resource limits
  - [ ] Storage backend specs 
- [ ] V2 Status API aggregates over multiple runtime indicators of health (Ring status, etc...)

alexandreLamarre:
Note : the alerting cluster driver interface will also be separate from the ops server API contract.

PR #1330, opened 2023-04-19
Alerting cluster crd

### LINKED ISSUE #1152: Standalone Alerting Controller CRD
- [ ] Separate Alerting Backend spec from Opni gateway CRD into its own AlertingCluster CRD
- [ ] Update Opni-Alerting's Kubernetes cluster drivers

alexandreLamarre:
Closes #1152


apis/core/v1beta1/alertingcluster_types.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Overrides for out-of-the box alerting specs
+	ApplicationSpec AlertingApplicationSpec `json:"workloads,omitempty"`
+}
+
+type AlertingApplicationSpec struct {
+	Replicas             *int32                            `json:"replicas,omitempty"`
+	ExtraArgs            []string                          `json:"extraArgs,omitempty"`
+	ExtraVolumes         []corev1.Volume                   `json:"extraVolumes,omitempty"`
+	ExtraVolumeMounts    []corev1.VolumeMount              `json:"extraVolumeMounts,omitempty"`
```
dbason, line 44: Not required for this iteration but a lot of operators combine the extra Volume and Volume mounts into a single object as a QoL feature because most of the time these are 1:1

pkg/resources/alerting/base.go
```diff
@@ -0,0 +1,387 @@
...
+		{
+			Name: "opni-alertmanager-data",
+			VolumeSource: corev1.VolumeSource{
+				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+					ClaimName: "opni-alertmanager-data",
+					ReadOnly:  false,
+				},
+			},
+		},
+	}
```
dbason, line 56: What happens if the user has multiple storage classes and they want to select one?  What if they don't have any storage classes?

pkg/resources/alerting/base.go
```diff
@@ -0,0 +1,387 @@
...
+				Name: "POD_IP",
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "status.podIP",
+					},
+				},
+			},
+			{
+				Name:  "USER",
+				Value: "syncer",
```
dbason, line 241: Nit this should be set as a constant and reused

pkg/resources/alerting/base.go
```diff
@@ -0,0 +1,387 @@
...
+				Name: "POD_IP",
+				ValueFrom: &corev1.EnvVarSource{
+					FieldRef: &corev1.ObjectFieldSelector{
+						FieldPath: "status.podIP",
+					},
+				},
+			},
+			{
+				Name:  "USER",
+				Value: "alerting",
```
dbason, line 180: This should probably be a constant

plugins/alerting/pkg/alerting/drivers/alerting_manager/cluster_driver.go
```diff
@@ -70,66 +48,57 @@ func NewAlertingManagerDriver(options AlertingManagerDriverOptions) (*AlertingMa
...
-	lg.Debug("got existing gateway")
-	mutator := func(gateway *corev1beta1.Gateway) {
-		gateway.Spec.Alerting.Replicas = conf.NumReplicas
-		gateway.Spec.Alerting.ClusterGossipInterval = conf.ClusterGossipInterval
-		gateway.Spec.Alerting.ClusterPushPullInterval = conf.ClusterPushPullInterval
-		gateway.Spec.Alerting.ClusterSettleTimeout = conf.ClusterSettleTimeout
-		gateway.Spec.Alerting.Storage = conf.ResourceLimits.Storage
-		gateway.Spec.Alerting.CPU = conf.ResourceLimits.Cpu
-		gateway.Spec.Alerting.Memory = conf.ResourceLimits.Memory
+func (a *AlertingClusterManager) InstallCluster(ctx context.Context, empty *emptypb.Empty) (*emptypb.Empty, error) {
```
dbason, line 70: I think it's probably worthwhile to separate the specific driver implementation (kubernetes and others) from the implementation of the gRPC interface.  @kralicky thoughts?

dbason (changes requested):

plugins/alerting/pkg/alerting/drivers/alerting_manager/cluster_driver.go
dbason: To expand on this the driver should be something the implementation of the gRPC abstraction uses rather than being the entirety of the gRPC interface.  I think there's value in separating these two concepts.

pkg/resources/alerting/base.go
alexandreLamarre: that's good point, I'll look into adding a storage class to the alerting cluster crd

plugins/alerting/pkg/alerting/drivers/alerting_manager/cluster_driver.go
alexandreLamarre: yeah agreed, this is part of the reason I added AlertOpsV2 spec to the epic this part of, to rework these issues : https://github.com/rancher/opni/issues/1153 . It is definitely not the best to couple the type contract for these definitions

Although the ops server is separated from the driver in a pretty thin encapsulation layer :

https://github.com/rancher/opni/blob/e48247bd6eaa78a2a2ffe96eff8d95ba5246cd6b/plugins/alerting/pkg/alerting/ops/ops.go

alexandreLamarre: Also since the epic is moving towards deploying cortex-alertmanager anyways, it might be more productive to rework the driver interface later when tackling https://github.com/rancher/opni/issues/1153

kralicky: yeah I agree @dbason, it would be best if the driver is part of the implementation of that server.

pkg/resources/alerting/base.go
alexandreLamarre: Ended up using a more generic approach of overloading the PVC template Spec

alexandreLamarre: Using the PVCSource from opster seems like a better approach here.

dbason (approved):

PR #1334, opened 2023-04-21
OTEL prometheus discovery controller : UX/consistency 

### LINKED ISSUE #1332: OTEL metric processing (UX)
Specifically for OTEL metric collection based on our custom prometheus CRD service discovery, within reason we want to : 
- [x] Make job names consistent with prometheus-agent
- [x] Make certain labels consistent with prometheus-agent
- [x] Existing opni grafana dashboards should work with OTEL capability

We also want to make sure which metrics are available for scraping to be compatible with the CRD definitions:
- [x] Mount secrets on demand for TLS communication based on promCRD definitions

alexandreLamarre:
Closes #1332 

With this PR we reach almost feature parity with Prometheus Operator, supporting :
- Relabeling metrics at scrape time
- Relabeling metrics after scrape time
- Service account auth for scrape targets
-  Authorization using tokens (Bearer, etc...) for scrape targets
- TLS auth for scrape targets
- Automatic discovery of required secrets : Prometheus operator checks the service monitor for local references (or other things) to config-maps/secrets and aggregates them into a mega secret. 
- Friendly job name (required for many exporter's recording/alerting rules)
- Compatible unique job names with Prometheus operator

<hr/>

Here's some Prometheus operator features that we don't support yet : 
- HTTP2 Prometheus scraping
- Auth subsets of prometheus HTTP client, such as OAUTH2|
- A lot of meta relabeling, as far as I know metrics exporters don't generally rely on these


pkg/resources/collector/workloads.go
```diff
@@ -22,9 +22,9 @@ const (
 	receiversKey       = "receivers.yaml"
 	mainKey            = "config.yaml"
 	aggregatorKey      = "aggregator.yaml"
-	collectorImageRepo = "ghcr.io/rancher-sandbox"
-	collectorImage     = "opni-otel-collector"
-	collectorVersion   = "v0.1.1-0.74.0"
+	collectorImageRepo = "docker.io/alex7285"
```
dbason, line 25: Is this the correct repo?

pkg/test/environment.go
```diff
@@ -873,7 +873,7 @@ func (t TestNodeConfig) MetricReceivers() []string {
 	res := []string{}
 	if t.Metrics.Enabled {
 		res = append(res, "prometheus/self")
-		if t.Metrics.Spec.HostMetrics {
+		if t.Metrics.Spec.HostMetrics != nil && *t.Metrics.Spec.HostMetrics {
```
dbason, line 876: Minor nit but it's probably cleaner to use something like
```go
if lo.FromPtrOr(t.Metrics.Spec.HostMetrics, false) {
```

dbason (changes requested):

pkg/resources/collector/workloads.go
alexandreLamarre: ah right need to release a new version, missed that

pkg/test/environment.go
alexandreLamarre: agreed, keep the nits coming

pkg/resources/collector/workloads.go
alexandreLamarre: https://github.com/rancher-sandbox/opni-otel-collector/pull/5, I know these PRs are overkill but it makes sense to me to document and sanity check changes to our otel distro

kralicky (changes requested):

kralicky (approved):

dbason (approved):

Issue #1136, opened 2023-03-06
[EPIC] Metric anomaly pattern matching


sanjay920:
## Issues:

---

## Summary: 
This enhancement proposal offers a new feature to match metrics to 2 types of known pre-defined anomaly chart patterns. For each time of the patterns, a grafana dashboard that visualize all anomaly metrics that match to the patterns will be dynamically generated for user to consume. This feature will make operators' life easier in the process of outage diagnosis.

The 2 types of pre-defined patterns:
<img width="1401" alt="Screen Shot 2023-02-22 at 1 54 28 PM" src="https://user-images.githubusercontent.com/4568163/220768433-82056fd1-494c-4afe-9822-77b0566fd99d.png">

## Use case: 
* Keeping track of all metrics in a multi-cluster setup is challenging. Subject matter experts could be responsible for hundreds of clusters and each cluster has tons of different metrics. Often times, SMEs are required to identify metrics that are important to a workload. Because only certain metrics are selected, there can be blind spots - as other metrics may capture details of the system overlooked by the metrics curated by SMEs. When an outage occurs, it's difficult to identify root-cause metrics to assist outage diagnosis. Users need tools to help them reduce the scope and identify anomaly metrics faster.

## Implementation details: 
![Flow chart](https://user-images.githubusercontent.com/4568163/220794014-5e4b70d2-94f5-451d-9e68-f5b27e1e2a24.png)

* UI -- Admin dashboard
	- Metric insights page: Add a button in the UI for user to manually invoke this feature.
	- Alert overview page: Attach a button to alerts for user to manually invoke this feature.
	- the UX: a table that displays all downstream clusters so user can easily select 1 or more of them to generate metric-analysis report.

* UI -- Dynamic Grafana dashboard: For each request of metric-analysis, generate 2 grafana dashboards for detected anomaly metrics with classified pattern types, 1 dashboard for metrics of Type-I patterns and another dashboard for metrics of Type-II patterns. 

* Metric-analysis service
	- HTTP server: serves backend APIs using FastAPI. APIs:
		* POST /request: post a request to invoke this feature, returns a random ID that can be used later. Spawns a background process to run the metric-analysis process.
		* GET /get_all_requests_id: get all the history requests posted, return all the ids.
		* GET /get_analysis_report_by_id: get an analysis report of given `id`.
		* GET /get_all_analysis_report: get all history reports
		* DELETE /delete_report_by_id: delete a report by id
		* DELETE /delete_all_reports: delete all history reports
[...]

PR #1340, opened 2023-04-24
Metric ai gateway plugin

tybalex:
This is the gateway plugin implementation of this [Epic](https://github.com/rancher/opni/issues/1136) .
a few concepts in this service:
* A job: A list of namespaces of a cluster to be watched, selected by user, similar to the watchlist we already have in log anomaly detection
* A job run: run metric analysis for a job at the requested timestamp.

What's included in this PR:
1. APIs to list/create/view/delete jobs:
    * ListJobs
    * CreateJob
    * DeleteJob
    * GetJob
2. APIs to list/run/view/delete job-run:
    * ListJobRuns
    * RunJob
    * DeleteJobRun
    * GetJobRunResult
3. APIs to create/delete grafanadashboard :
    * CreateGrafanaDashboard
    * DeleteGrafanaDashboard
4. Other helper APIs that UI needs:
    * ListClusters
    * ListNamespaces


AmartC (commented):
Can you please add a description of what this PR is doing. Additionally, fix the build errors.

plugins/aiops/pkg/gateway/metricai.go
```diff
@@ -0,0 +1,363 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const serverUrl = "http://metric-ai-service:8090" // URL of the python metric-ai service
+const jobRunDelimiter = "="                       // delimiter that splits jobid and suffix, in order to save in natsKV
+const dashboardNamePrefix = "metricai-"
+
```
dbason, line 25: Stylistically I prefer to group these up, i.e
```go
const (
	serverUrl = "http://metric-ai-service:8090" // URL of the python metric-ai service
	jobRunDelimiter = "="                       // delimiter that splits jobid and suffix, in order to save in natsKV
	dashboardNamePrefix = "metricai-"
)
```

plugins/aiops/pkg/gateway/metricai.go
```diff
@@ -0,0 +1,363 @@
...
+func (p *AIOpsPlugin) ListClusters(ctx context.Context, _ *emptypb.Empty) (*metricai.MetricAIIdList, error) {
+	// For the UI to list clusters. Returns cluster_id
+
+	timeout := 10 * time.Second
+	ctxca, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	httpClient := &http.Client{Timeout: timeout}
+	// make the http request with context
+	req, err := http.NewRequestWithContext(ctxca, "GET", serverUrl+"/get_users", nil)
```
dbason, line 109: Rather than `"GET"` should use `http.MethodGet`

plugins/aiops/pkg/gateway/metricai.go
```diff
@@ -0,0 +1,363 @@
...
+	return &metricai.MetricAIAPIResponse{SubmittedTime: time.Now().String(), Description: dashboardJson, Status: "Success"}, nil
+}
+
+func (p *AIOpsPlugin) ListClusters(ctx context.Context, _ *emptypb.Empty) (*metricai.MetricAIIdList, error) {
+	// For the UI to list clusters. Returns cluster_id
+
+	timeout := 10 * time.Second
+	ctxca, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	httpClient := &http.Client{Timeout: timeout}
```
dbason, line 107: Probably won't be an issue but for scaling I think it's probably better to create the httpClient in the struct and reuse the same one.

plugins/aiops/pkg/gateway/metricai.go
```diff
@@ -0,0 +1,363 @@
...
+		if errors.Is(err, nats.ErrKeyNotFound) {
+			return nil, &RequestError{
+				StatusCode: 503,
+				Err:        errors.New("The jobId to delete doesn't exist"),
+			}
+		}
+		return nil, status.Errorf(codes.Internal, "Failed to DeleteJob for metricAI: %v", err)
+	}
+	metricAIKeyValue.Delete(jid)
+	return &metricai.MetricAIAPIResponse{Status: "Success", Description: "The JobId key :" + jid + " is deleted"}, nil
```
dbason, line 286: Nit: rather than using concatenation we should use sprintf e.g
```go
Description: fmt.Sprintf("The JobId key :%s is deleted", jid)
```

dbason (changes requested):
Some minor stuff.  I've only commented on single incidences but there are more within the code.

More generally I think we need to abstract some of the nats KV setup out of the plugins and provide an interface for getting/creating a KV bucket.  We probably also need to clarify the impact of using multiple buckets vs using KVs within a single bucket.  This isn't something we need to fix in this PR but we should add cards for this.

tybalex:
> Some minor stuff. I've only commented on single incidences but there are more within the code.
> 
> More generally I think we need to abstract some of the nats KV setup out of the plugins and provide an interface for getting/creating a KV bucket. We probably also need to clarify the impact of using multiple buckets vs using KVs within a single bucket. This isn't something we need to fix in this PR but we should add cards for this.

Agreed. I do notice some other  gateway plugins use natsKV too, such as metrics and alerts, and there exists different implementations for their own usage. In general it's definitely a good idea to provide an interface for this so it's easier to maintain everything.

Regard to the usage of using multiple buckets, It would be decent if we can clarify the impact of doing so. But I  think as long as we don't overwhelm it -- aka creating tons of buckets -- we should be fine. In my case I am using 2 buckets: 1 for manage the metadata for each job, the other for each job_run(includes the generated granfana dashboard JSON string).

plugins/aiops/pkg/gateway/metricai.go
tybalex: updated

plugins/aiops/pkg/gateway/metricai.go
tybalex: good point. fixed

plugins/aiops/pkg/gateway/metricai.go
tybalex: fixed all the similar messages

plugins/aiops/pkg/gateway/metricai.go
tybalex: fixed.

plugins/aiops/pkg/gateway/metricai.go
```diff
@@ -0,0 +1,365 @@
...
+			},
+			Spec: grafanav1alpha1.GrafanaDashboardSpec{
+				Json: dashboardJson,
+			},
+		},
+	}
+	for _, dashboard := range grafanaDashboards {
+		err := p.k8sClient.Create(ctx, dashboard)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Error Creating Dashboard for metricAI: %v", err)
```
alexandreLamarre, line 70: "%v" in Errorf, depending on the error, may cast it to bytes or a pointer, `%s` or `%w` is safer for logging a readable error

plugins/aiops/pkg/gateway/metricai.go
```diff
@@ -0,0 +1,365 @@
...
+	ctxca, cancel := context.WithTimeout(ctx, 10*time.Second)
+	defer cancel()
+	metricAIKeyValue, err := p.metricAIJobKv.GetContext(ctxca)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to ListJobRuns for metricAI: %v", err)
+	}
+	jid := strings.ToLower(jobRequest.JobId)
+
+	if jid == "" { // id can't be empty
+		return nil, &RequestError{
```
alexandreLamarre, line 232: Using this struct definition won't convert codes in the GRPC response, for that we should prefer using grpc `codes` package like in the other apis in this file. 

stylistic nit : for input validation we typically use "github.com/rancher/opni/pkg/validation" and implement the Validator interface, see for example https://github.com/rancher/opni/blob/main/pkg/apis/control/v1/validate.go

alexandreLamarre (dismissed):
Optional nits to consider

plugins/aiops/pkg/gateway/metricai.go
tybalex: fixed

plugins/aiops/pkg/gateway/metricai.go
tybalex: good point. Added code using the validator interface.

tybalex:
Close this PR as this branch has been merged into the other branch `metric-service` which will includes the python service implementation and integration test.

PR #1364, opened 2023-04-30
Fix delete of old Opensearch data

### LINKED ISSUE #1353: Delete data does not remove data from warm indices
When ISM rolls over the index it forces it into read only mode.  This stops the delete_by_query from deleting data from warm indices.

@AmartC FYI

dbason:
Fixes #1353

### DIFF
--- a/plugins/logging/pkg/opensearchdata/delete.go
+++ b/plugins/logging/pkg/opensearchdata/delete.go
@@ -12,6 +12,9 @@ import (
 	"github.com/tidwall/sjson"
 )
 
+// TODO: Tracking of external system task state should be generalized and moved
+// to the generic task controller
+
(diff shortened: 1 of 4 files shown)


plugins/logging/pkg/opensearchdata/delete.go
```diff
@@ -148,3 +148,19 @@ func (m *Manager) DeleteTaskStatus(ctx context.Context, id string, readyFunc ...
 
 	return DeleteFinished, nil
 }
+
+func (m *Manager) DeleteKVKey(ctx context.Context, id string) error {
```
kralicky, line 152: this seems a bit too hacky :stuck_out_tongue:

kralicky (dismissed):

kralicky (approved):

PR #1368, opened 2023-05-01
One time use token

### LINKED ISSUE #1366: Implement one time use bootstrap tokens
Currently bootstrap tokens can be used infinite times within their TTL.  It would be very useful to have a token that would only be used once.  This issue is to investigate what is required for this feature.

dbason:
Fixes #1366


alexandreLamarre (approved):

pkg/apis/core/v1/core.proto
```diff
@@ -20,6 +20,7 @@ message BootstrapTokenMetadata {
   map<string, string> labels = 4;
   repeated TokenCapability capabilities = 5;
   string resourceVersion = 6;
+  bool onetime = 7;
```
kralicky, line 23: This should be implemented via a `maxUsages` field. Since usage counts are already tracked, that would integrate nicely with it.

pkg/bootstrap/server_v2.go
```diff
@@ -153,13 +153,21 @@ func (h *ServerV2) Auth(ctx context.Context, authReq *bootstrapv2.BootstrapAuthR
...
+	} else {
+		_, err = h.storage.UpdateToken(ctx, token.Reference(),
+			storage.NewCompositeMutator(
+				storage.NewIncrementUsageCountMutator(),
+			),
+		)
+		if err != nil {
+			return nil, status.Error(codes.Internal, fmt.Sprintf("error incrementing usage count: %v", err))
+		}
 	}
```
kralicky, line 171: This should not have to be changed, the storage backend should handle expiring the token when it has reached its usage count limit (see above comment)

pkg/storage/jetstream/token_store.go
```diff
@@ -26,6 +26,7 @@ func (s *JetStreamStore) CreateToken(_ context.Context, ttl time.Duration, opts
 		UsageCount:   0,
 		Labels:       options.Labels,
 		Capabilities: options.Capabilities,
+		Onetime:      options.OneTime,
```
kralicky, line 29: Need to implement this in all storage backends

kralicky (changes requested):

pkg/opni/commands/tokens.go
```diff
@@ -56,6 +58,7 @@ func BuildTokensCreateCmd() *cobra.Command {
 	}
 	tokensCreateCmd.Flags().StringVar(&ttl, "ttl", "300s", "Time to live")
 	tokensCreateCmd.Flags().StringSliceVar(&labels, "labels", []string{}, "Labels which will be auto-applied to any clusters created with this token")
+	tokensCreateCmd.Flags().IntVar(&maxUsages, "max-usages", 0, "Maximum number of times this token can be used")
```
alexandreLamarre, line 61: Since this is user facing, we should clarify here that 0 is unlimited

kralicky (approved):

PR #1392, opened 2023-05-05
Import error

### LINKED ISSUE #1373: Data import does not always report failed imports appropriately
Sometimes a data import will fail but it will still be reported as `running`. Error messages are shown when checking its status and in the logs, but the import is not.

```
2023-05-02T18:25:26Z INFO tasks task/task.go:71 import running {"id": "demo"}
2023-05-02T18:25:27Z DEBUG tasks task/task.go:69 pushed 23326 bytes to remote write {"id": "demo"}
...
2023-05-02T18:25:28Z ERROR tasks task/task.go:75 failed to push to remote write: rpc error: code = Canceled desc = context canceled {"id": "demo"}
```

```
 CLUSTER                               NAME   ENDPOINT           LAST READ                      STATE    MESSAGE                                                                  
 7f67db15-9534-4bae-9ba8-53618b130c27  demo2  https://something  2023-05-02 17:30:07 +0000 UTC  running  failed to push to remote write: rpc error: code = context cancelled
```

### LINKED ISSUE #1391: Import tool should be smarter when handling failed pushes
Currently, when the import tool fails to push data to the remote, it will simply log the failure before continuing with the next import time range. This will cause data to be dropped if an internet connection is dropped, in this case we should retry the push with an exponential backoff. On the other hand, if we encounter a non-recoverable issue such as a canceled context we should stop the import and report it as failed.

We need to implement a more intelligent error handling. We should check if an error is recoverable and continue with backoff, otherwise just cancel the import.

joshmeranda:
Improves how we handle errors in the data import.

closes #1391 
closes #1373

### DIFF
--- a/plugins/metrics/pkg/agent/runner.go
+++ b/plugins/metrics/pkg/agent/runner.go
@@ -184,6 +184,10 @@ func (tr *taskRunner) doPush(ctx context.Context, writeRequest *prompb.WriteRequ
 				_, err = promClient.Push(ctx, writeRequest)
 			})
 
+			if err == nil {
+				return nil
+			}
+
 			if !isRecoverable(err) {
 				return fmt.Errorf("failed to push to remote write: %w", err)
 			}
--- a/plugins/metrics/pkg/agent/runner.go
+++ b/plugins/metrics/pkg/agent/runner.go
@@ -216,15 +216,17 @@ func (tr *taskRunner) OnTaskRunning(ctx context.Context, activeTask task.ActiveT
 				Timeseries: dereferenceResultTimeseries(result.Timeseries),
 			}
 
+			var pushErr error
 			tr.remoteWriteClient.Use(func(remoteWriteClient remotewrite.RemoteWriteClient) {
 				promClient := remotewrite.AsPrometheusRemoteWriteClient(remoteWriteClient)
 				if _, err := promClient.Push(context.Background(), &writeRequest); err != nil {
-					activeTask.AddLogEntry(zapcore.ErrorLevel, fmt.Sprintf("failed to push to remote write: %s", err))
-					return
+					pushErr = fmt.Errorf("failed to push to remote write: %w", err)
+					activeTask.AddLogEntry(zapcore.ErrorLevel, pushErr.Error())
 				}
-
-				activeTask.AddLogEntry(zapcore.DebugLevel, fmt.Sprintf("pushed %d bytes to remote write", writeRequest.Size()))
 			})
+			if pushErr != nil {
+				return pushErr
+			}
 
 			progress.Current = uint64(nextEnd - progressDelta)
 			activeTask.SetProgress(progress)
--- a/plugins/metrics/pkg/agent/runner.go
+++ b/plugins/metrics/pkg/agent/runner.go
@@ -175,7 +175,7 @@ func (tr *taskRunner) doPush(ctx context.Context, writeRequest *prompb.WriteRequ
 	for {
 		select {
 		case <-expbackoff.Done():
-			return fmt.Errorf("failed to push data after exponential backoff")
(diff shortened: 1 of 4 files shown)


plugins/metrics/pkg/agent/runner.go
```diff
@@ -148,6 +169,32 @@ func (tr *taskRunner) OnTaskPending(_ context.Context, _ task.ActiveTask) error
...
+			return fmt.Errorf("failed to push data after exponential backoff")
+		case <-expbackoff.Next():
+			var err error
+
+			tr.remoteWriteClient.Use(func(remoteWriteClient remotewrite.RemoteWriteClient) {
+				promClient := remotewrite.AsPrometheusRemoteWriteClient(remoteWriteClient)
+				_, err = promClient.Push(ctx, writeRequest)
+			})
+
+			if !isRecoverable(err) {
```
alexandreLamarre, line 187: Just wondering what the behaviour is here when the gateway/cortex restarts? is it context cancelled

alexandreLamarre (approved):

plugins/metrics/pkg/agent/runner.go
kralicky: If ctx is canceled, the backoff policy will be canceled and the task will go into the failed state.

alexandreLamarre (dismissed):

plugins/metrics/pkg/agent/runner.go
kralicky: sorry, to clarify, this appears to be incorrect logic.

kralicky (changes requested):

plugins/metrics/pkg/agent/runner.go
joshmeranda: Shoot you're right. We need to make the distinction between the push context failing and the task context failing. I'll push a fix in a bit.

test/plugins/metrics/metrics_suite_test.go
```diff
@@ -4,12 +4,20 @@ import (
...
 
 	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/ginkgo/v2/types"
 	. "github.com/onsi/gomega"
 	_ "github.com/rancher/opni/pkg/test/setup"
 	_ "github.com/rancher/opni/plugins/metrics/test"
 )
 
 func TestMetrics(t *testing.T) {
+	suiteConfig := types.NewDefaultSuiteConfig()
```
kralicky, line 14: did you mean to leave these in here?

plugins/metrics/pkg/agent/runner.go
```diff
@@ -148,6 +169,43 @@ func (tr *taskRunner) OnTaskPending(_ context.Context, _ task.ActiveTask) error
 	return nil
 }
 
+func (tr *taskRunner) doPush(ctx context.Context, writeRequest *prompb.WriteRequest) error {
+	expbackoff := tr.backoffPolicy.Start(ctx)
+
+	for {
```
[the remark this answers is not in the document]
joshmeranda: @kralicky  Is something like this what you're looking for?

```
select {
case <-ctx.Done():
    return ctx.Err()
default:
}
```

This should do what you're asking and still passes the new tests

test/plugins/metrics/metrics_suite_test.go
joshmeranda: nope, I'll take those out thanks for catching

plugins/metrics/pkg/agent/runner.go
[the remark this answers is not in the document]
joshmeranda: fixed and pushed

kralicky (approved):

PR #1396, opened 2023-05-10
Add support for nested message fields in http path templates

kralicky:
Closes https://github.com/rancher/opni/issues/1397


pkg/management/extensions.go
```diff
@@ -493,10 +494,39 @@ func unknownServiceHandler(director StreamDirector) grpc.StreamHandler {
...
+			}
+			return fmt.Errorf("field path %q is invalid: %q is not a message", k, fd.GetFullyQualifiedName())
+		}
+		msg := fd.GetMessageType()
+		var subMsg *dynamic.Message
+		if fld := containingMsg.GetField(fd); reflect.ValueOf(fld).IsNil() {
+			subMsg = dynamic.NewMessage(msg)
+			containingMsg.SetField(fd, subMsg)
+		} else {
+			subMsg = fld.(*dynamic.Message)
```
alexandreLamarre, line 522: for consistency with our other impls and to be on the safe side we should probably type check `msg, ok := fld.(*dynamic.Message)`

alexandreLamarre (changes requested):

alexandreLamarre (approved):

PR #1398, opened 2023-05-10
Support Agent authentication

dbason:
This PR containers the following pieces of work:

1. Minor refactor of ident provider loading
2. Remove deprecated Subjects method for certs
3. Add a token bucket ratelimiter to gRPC streams
4. Add a ping service to the gateway stream
5. Create a command to bootstrap a support agent


pkg/bootstrap/server_v2.go
```diff
@@ -136,7 +136,12 @@ func (h *ServerV2) Auth(ctx context.Context, authReq *bootstrapv2.BootstrapAuthR
 	kr := keyring.New(keyring.NewSharedKeys(sharedSecret))
 
 	tokenLabels := maps.Clone(bootstrapToken.GetMetadata().GetLabels())
-	delete(tokenLabels, corev1.NameLabel)
+
+	// if the token is not a one-time token, remove the name label
+	if bootstrapToken.GetMetadata().GetMaxUsages() != 1 {
+		delete(tokenLabels, corev1.NameLabel)
+	}
```
kralicky, line 143:
```suggestion
	// if the token is not a one-time token, remove the name label
	if bootstrapToken.GetMetadata().GetMaxUsages() - bootstrapToken.GetMetadata().GetUsageCount() > 1 {
		delete(tokenLabels, corev1.NameLabel)
	}
```

pkg/config/config.go
```diff
@@ -97,6 +101,39 @@ func FindConfig() (string, error) {
...
+				f.Close()
+				return p, nil
+			}
+		}
+	}
+
+	return "", ErrConfigNotFound
+}
+
+func FindSupportConfig() (string, error) {
```
kralicky, line 125: This seems redundant (or is it unused?) given the changes to FindConfig above

go.mod
```diff
@@ -55,6 +56,7 @@ require (
 	github.com/jaypipes/ghw v0.10.0
 	github.com/jedib0t/go-pretty/v6 v6.4.6
 	github.com/jhump/protoreflect v1.14.1
+	github.com/juju/ratelimit v1.0.2
```
kralicky, line 58: this module is LGPL, but https://pkg.go.dev/golang.org/x/time/rate appears to have the same functionality (and it's an existing dependency too)

kralicky (changes requested):

pkg/supportagent/supportagent.go
kralicky, line 1: I would highly recommend basic integration with a library such as https://github.com/99designs/keyring for keyring storage

pkg/trust/flags.go
```diff
@@ -53,11 +53,11 @@ func BuildConfigFromFlags(flags *pflag.FlagSet) (*StrategyConfig, error) {
 			if err != nil {
 				return nil, err
 			}
-			cert, err := util.ParsePEMEncodedCert(data)
+			certchain, err := util.ParsePEMEncodedCertChain(data)
```
kralicky, line 56: This appears incorrect - each file should only contain a single cert, or if needed we can parse the cert chain and extract the root certificate. Either way we only want to add root ca certs to this list.

dbason: I'm not sure that makes sense.  We need to handle the case where a user is passing in a list of roots (like system roots, or a similar file)

pkg/config/config.go
dbason: I'll back out the changs in the FindConfig function as I wanted to be more restrictive here.

pkg/bootstrap/server_v2.go
dbason: MaxUsages of 0 is unlimited usages so this change will not be valid.

kralicky: resolved offline, original logic is correct

test/integration/gateway_ratelimit_test.go
```diff
@@ -0,0 +1,49 @@
...
+				go func() {
+					_, connErrC := environment.NewStreamConnection([]string{fingerprint})
+					for err := range connErrC {
+						errC <- err
+					}
+				}()
+			}
+			Eventually(errC).WithPolling(5 * time.Millisecond).Should(Receive(And(
+				Not(BeNil()),
+				MatchError(ContainSubstring("is unable to handle the request, please retry later")),
```
kralicky, line 45: Should check the grpc status code here. You can use `testutil.MatchStatusCode` for this.

kralicky (approved):

kralicky (approved):

PR #1403, opened 2023-05-11
return a default config if no alerting cluster crd is found

alexandreLamarre:


alexandreLamarre:
Issue was resolved in a previous PR.

PR #1405, opened 2023-05-12
Fix api semantics of the http body field config to be consistent with existing usage

kralicky:


pkg/management/extensions.go
```diff
@@ -316,7 +316,44 @@ func newHandler(
 				status.Errorf(codes.InvalidArgument, err.Error()))
 			return
 		} else if len(body) > 0 {
-			if err := reqMsg.UnmarshalMergeJSON(body); err != nil {
+			var err error
+			bodyFieldPath := rule.GetHttp().GetBody()
+			if bodyFieldPath == "" {
+				bodyFieldPath = "*"
+			}
```
alexandreLamarre, line 323: Can you clarify this? Does this mean all protos that do not have the `google.http.api.body` options set will have it implicitly set as `google.http.api.body : *`?

pkg/management/extensions.go
```diff
@@ -316,7 +316,44 @@ func newHandler(
...
+					bodyField := reqMsg.GetField(fd)
+					var dynamicBodyField *dynamic.Message
+					if reflect.ValueOf(bodyField).IsNil() {
+						dynamicBodyField = dynamic.NewMessage(fd.GetMessageType())
+						reqMsg.SetField(fd, dynamicBodyField)
+					} else {
+						var ok bool
+						dynamicBodyField, ok = bodyField.(*dynamic.Message)
+						if !ok {
+							panic("bug: request is not a *dynamic.Message")
```
alexandreLamarre, line 344: I can see it both ways but returning `InternalServerError` here might make sense here

pkg/management/extensions.go
```diff
@@ -316,7 +316,44 @@ func newHandler(
...
+						}
+					}
+					err = dynamicBodyField.UnmarshalMergeJSON(body)
+				} else {
+					if err := decodeAndSetField(reqMsg, bodyFieldPath, string(body)); err != nil {
+						runtime.HTTPError(ctx, mux, outboundMarshaler, w, req,
+							status.Errorf(codes.InvalidArgument, "failed to decode body field %q: %v", bodyFieldPath, err))
```
alexandreLamarre, line 351: `decodeAndSetField` can return errors related to unsupported types, it might be worth differentiating those errors here as something other than `InvalidArgument`

alexandreLamarre (changes requested):

pkg/management/extensions.go
kralicky: Yeah, this matches the logic in grpc-gateway (but notably does not match the logic in upstream google apis)

pkg/management/extensions.go
kralicky: good catch, this should be handled differently

kralicky: actually this looks correct, upon further inspection the dynamic package will always unmarshal by creating nested dynamic messages.

alexandreLamarre: yeah i agree this control flow is clearly unrecoverable, just a question around considering error semantics

pkg/management/extensions.go
alexandreLamarre: this seems side-effecty, are there some edge cases we can add to the tests? also we should verify this doesn't effect consumers of the api like the dashboard

kralicky: To clarify, this is the same as the previous implementation. from grpc-gateway:
```
func (b binding) GetBodyFieldPath() string {
	if b.Body != nil && len(b.Body.FieldPath) != 0 {
		return b.Body.FieldPath.String()
	}
	return "*"
}
```

alexandreLamarre (approved):

PR #1421, opened 2023-05-18
Metrics otel collector : endpoint ordering edge cases and watch improvements

### LINKED ISSUE #1416: OpenTelemetry metrics discovery : no route to host
Cortex service monitor targets are incorrectly discovered. This bug may apply to more generic cases.

<details>
<pre><code>
2023-05-17T15:47:19.806Z	debug	scrape/scrape.go:1351	Scrape failed	{"kind": "receiver", "name": "prometheus/discovered", "data_type": "metrics", "scrape_pool": "serviceMonitor/opni/cortex-store-gateway/0", "target": "https://10.0.28.102:8080/metrics", "error": "Get \"https://10.0.28.102:8080/metrics\": dial tcp 10.0.28.102:8080: connect: no route to host"}
</pre></code>
</details>

alexandreLamarre:
Fixes #1416


controllers/core_collector_controller.go
```diff
@@ -70,14 +76,62 @@ func (r *CoreCollectorReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		).
+		// We need to watch almost all pod events here, but instead of queuing reconciles
+		// for each event, we "hack" a batch mechanism for potential pod events
+		// using a periodic reconciliation channel.
+		Watches(
+			&source.Channel{
+				Source: periodicRecChan,
+			},
+			requestMapper,
+		).
```
dbason, line 134: As discussed on slack a better mechanism is to use the `RequeueAfter` field of `reconcile.Result` to force this to requeue after 60 seconds.

A couple of alternative approaches might also be:
- Use event filtering to filter out some of the pod update events
- Use a ratelimiter that's more aggressive that the 100, 10qps default token bucket.

These can be added to the controller with the `WithEventFilter` and `WithOptions` functions.

dbason (changes requested):

controllers/core_collector_controller.go
alexandreLamarre: Seems `WithEventFilter` is equivalent to (globally) setting predicates on watched resources.

I tried using the following predicates on pod events, but for whatever reason it wasn't catching updates to pods that were managed by other controllers

```
		Watches(
			&source.Kind{Type: &corev1.Pod{}},
			requestMapper,
			builder.WithPredicates(
				predicate.Or(
					predicate.GenerationChangedPredicate{},
					predicate.ResourceVersionChangedPredicate{},
					predicate.NewPredicateFuncs(
						func(obj client.Object) bool {
							return true
						},
					),
				),
			),
		).		
```

My fallback is to re-queue every minute when metrics is enabled, but if we can find a set of predicates that work on pods we can probably bump up that re-queue timer to 15minutes, as a safeguard to catch updates

dbason (approved):

kralicky (dismissed):

kralicky (approved):

PR #1545, opened 2023-06-30
Add the typescript service generator

codyrancher:
Run:
- `mage build:typescriptServiceGenerator`
- `mage generate:protobufTypescript`

To generate the typescript service files. It does assume that yarn is installed globally.


kralicky:
Looks like `yarn build:service-generator` isn't writing bin/service-generator. 
image

magefiles/targets/generate_proto.go
```diff
@@ -83,6 +85,19 @@ func (Generate) ProtobufPython(ctx context.Context) error {
 	return nil
 }
 
+// Build the typescript service generator plugin
+func (Generate) TypescriptServiceGenerator() error {
```
kralicky, line 89: This should be part of the `Build` namespace, and ProtobufTypescript should depend on it via `mg.Deps(Build.TypescriptServiceGenerator)`

web/service-generator/src/http_pb.ts
kralicky, line 1: `generated/google/api/http_pb.ts` should be added to git and imported from the code generator instead of having a duplicate version here, now that we don't need to "bootstrap" it anymore.

web/.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
-dist
\ No newline at end of file
+dist
+pkg/opni/generated
```
kralicky, line 3: We should version control generated code (unless it causes problems with node or something)

kralicky (changes requested):

web/service-generator/src/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    });
+  });
+}
+
+function printMethod(f: GeneratedFile, method: DescMethod) {
+  const m = findCustomMessageOption(method, 72295728, HttpRule);
+  const input = f.import(method.input);
+  const output = f.import(method.output);
+
+  f.print(`export async function ${ method.name }(input:`, input, `): Promise<`, output, `> {
```
kralicky, line 30: The generated code preamble comments should be added here to match protoc-gen-es

web/.gitignore
codyrancher: Node shouldn't have an issue with it and I don't really have a strong opinion on it. In the past I've never checked in generated code and some of the arguments I've heard have been:
- Repos are mostly for tracking history and checking in the generated code makes the history less clear since the real change is in the generator.
- You're not losing anything by just checking in just the codegen since it should be idempotent and you can always generate again.
- Generated code isn't intended for reading

I'm fine to check it in but I'd like to heard why just since I've never done it.

magefiles/targets/generate_proto.go
codyrancher: Done.

Is there a better way to execute a command in a different directory?
```
        webPath := path.Join(cwd, "web")
	os.Chdir(webPath)

	runError := sh.RunV("yarn", "build:service-generator")
	os.Chdir(cwd)
```

web/service-generator/src/index.ts
codyrancher: Done

web/service-generator/src/http_pb.ts
codyrancher: I'll wait on this until we talk about the checked in code since I think it might affect where we keep this file.

web/.gitignore
kralicky: It is standard practice in the go ecosystem because go modules pull source code directly from git and without the generated code, the module won't be able to compile.

kralicky: For the typescript code it probably doesn't make much of a difference. Although I would say we should at least check in http_pb.ts to avoid needing to bootstrap the code generator every time.

magefiles/targets/generate_proto.go
kralicky: Yes, but not using sh.Run. You would have to use exec.Command directly.

web/.gitignore
codyrancher: Added http_pb.ts.

web/service-generator/src/http_pb.ts
codyrancher: Done

magefiles/targets/generate_proto.go
codyrancher: Updated with exec.Command. Let me know if I can improve further.

PR #1551, opened 2023-07-04
Scaling opni alerting as a microservice OEP

alexandreLamarre:
Reduces complexity of gateway and improves resiliency of individual alerting component failures.


dbason:
Will each operation be a separate service in the initial implementation?  What are the latency impacts if any of having each separated?  What are the failure modes if one of the microservices goes down?

Issue #1415, opened 2023-05-17
Data import Sending too large packets


joshmeranda:
Data import is sending packets too large for the totem gateway causing the gateway server to fail and stop the import.

```
2023-05-17T15:15:06Z DEBUG plugin.metrics.cortex-admin cortex/admin.go:189 handling query {"query": "{__name__=~\".+\"}"}
2023-05-17T15:15:53Z ERROR plugin.logging.logging-backend backend/metadata.go:38 opensearch client not set, waiting
2023-05-17T15:15:54Z INFO plugin.modeltraining gateway/system.go:67 waiting for k8s object
2023-05-17T15:16:03Z INFO plugin.metrics.metrics-backend backend/remoteread.go:60 added new target {"cluster": "0d2b9469-99ad-4c52-8567-3d6bf19da883", "target": "demo", "capability": "metrics"}
2023-05-17T15:16:08Z DEBUG gateway.delegate gateway/delegate.go:91 delegating rpc request {"target": "0d2b9469-99ad-4c52-8567-3d6bf19da883", "request": "/remoteread.RemoteReadAgent/GetTargetStatus"}
2023-05-17T15:16:19Z DEBUG gateway.delegate gateway/delegate.go:91 delegating rpc request {"target": "0d2b9469-99ad-4c52-8567-3d6bf19da883", "request": "/remoteread.RemoteReadAgent/Start"}
2023-05-17T15:16:19Z INFO plugin.metrics.metrics-backend backend/remoteread.go:262 target started {"cluster": "0d2b9469-99ad-4c52-8567-3d6bf19da883", "capability": "metrics", "target": "demo"}
2023-05-17T15:16:19Z DEBUG gateway.delegate gateway/delegate.go:91 delegating rpc request {"target": "0d2b9469-99ad-4c52-8567-3d6bf19da883", "request": "/remoteread.RemoteReadAgent/GetTargetStatus"}
2023-05-17T15:16:20Z DEBUG gateway.delegate gateway/delegate.go:91 delegating rpc request {"target": "0d2b9469-99ad-4c52-8567-3d6bf19da883", "request": "/remoteread.RemoteReadAgent/GetTargetStatus"}
2023-05-17T15:16:20.530Z WARN totem.gateway-server server.go:222 stream exited with error {"error": "rpc error: code = ResourceExhausted desc = grpc: received message larger than max (15349392 vs. 4194304)"}
[...]

PR #1552, opened 2023-07-05
Import chunks

joshmeranda:
fixes https://github.com/rancher/opni/issues/1415


plugins/metrics/pkg/agent/runner.go
```diff
@@ -195,18 +191,25 @@ func (tr *taskRunner) doPush(ctx context.Context, writeRequest *prompb.WriteRequ
...
-				zap.Error(err),
-			).Error("failed to push to remote write, retrying...")
 		}
 	}
 }
 
 func (tr *taskRunner) OnTaskRunning(ctx context.Context, activeTask task.ActiveTask) error {
+	limit := WriteLimit{
+		GrpcMaxBytes:             4194304,
+		CortexIngestionRateLimit: 2500,
```
kralicky, line 211: The default rate limit is 25000 samples/s, not 2500. That said, the agent should probably not need to worry about ingestion rate limits, and just handle the 429 status appropriately.

plugins/metrics/pkg/agent/runner.go
```diff
@@ -274,9 +275,21 @@ func (tr *taskRunner) OnTaskRunning(ctx context.Context, activeTask task.ActiveT
...
+			}
+
+			if len(chunkedRequests) > 1 {
+				activeTask.AddLogEntry(zapcore.DebugLevel, fmt.Sprintf("split write request into %d chunks", len(chunkedRequests)))
+			}
+
+			for i, request := range chunkedRequests {
+				if err := tr.doPush(ctx, request); err != nil {
+					activeTask.AddLogEntry(zapcore.ErrorLevel, err.Error())
+					return err
```
kralicky, line 290: should there be additional error handling logic here? e.g. handling 429

plugins/metrics/pkg/agent/chunks.go
kralicky, line 1: Can you add a memory benchmark test for this code?

kralicky (changes requested):

plugins/metrics/pkg/agent/runner.go
joshmeranda: Ooops, typo. I think in the case where a write request has more samples than can fit in the ingestion rate limit we do need to worry about the limits. If the request itself is too large to fit within that window we'll just keep entering the backoff.

plugins/metrics/pkg/agent/runner.go
joshmeranda: I think the current error handling backoff logic in `doPush` should already be sufficient. We could make that backoff configurable, but I think that level of configuration for the user likely isn't necessary.

plugins/metrics/pkg/agent/runner.go
joshmeranda: We should make the limit values configurable though, and part of the import spec.

plugins/metrics/pkg/agent/runner.go
```diff
@@ -195,18 +191,25 @@ func (tr *taskRunner) doPush(ctx context.Context, writeRequest *prompb.WriteRequ
 			default:
 			}
 
-			if !isRecoverable(err) {
+			switch {
+			case strings.Contains(err.Error(), "context cancelled"),
```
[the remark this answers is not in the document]
joshmeranda: For the cause behind this specific error handling, we could probably even take out this check for `context cancelled` since that specific check was kinda-sorta just a workaround for this bug.

plugins/metrics/pkg/agent/runner.go
kralicky: That's something I am working on right now, you'll be able to configure limits globally and also per-tenant as part of the cortexops api.

plugins/metrics/pkg/agent/chunks.go
```diff
@@ -0,0 +1,128 @@
...
+type WriteLimit struct {
+	// GrpcMaxBytes is the maximum siz of a message that can be sent over the grpc connection.
+	GrpcMaxBytes int
+
+	// CortexIngestionRateLimit is the maximum number of samples that can be sent to cortex per second.
+	CortexIngestionRateLimit int
+}
+
+func splitNChunks[T any](a []T, n int) ([][]T, error) {
+	if n == 0 {
```
jan-law, line 19: What happens if `n == 1`?

joshmeranda: when `n == 1` the populating loop below will iterate once, then `end` will exceed the length of the given slice, and finish. `splitNChunks([]int{0, 1, 2, 3}, 1) == [][]int{{0, 1, 2, 3}}`

kralicky (approved):

Issue #1413, opened 2023-05-17
Agent binary patching flake


alexandreLamarre:
Something caused a gateway patch sent to an agent to delete the metrics and logging plugins on the agent which (obviously) results in no logs or metrics being sent

<details>
<pre><code>
2023-05-17T14:06:19Z INFO commands/agent_v2.go:65 using config file {"path": "/etc/opni/config.yaml"}
2023-05-17T14:06:19Z DEBUG agent v2/agent.go:127 using log level: debug
2023-05-17T14:06:19Z INFO agent v2/agent.go:213 loaded existing keyring
2023-05-17T14:06:19Z DEBUG keyring machinery/keys.go:53 loaded ephemeral key {"path": "/run/opni-agent/keyring/session-attribute.json", "usage": "auth", "labels": {"opni.io/session-attribute":"local"}}
2023-05-17T14:06:19Z INFO agent v2/agent.go:500 attempting to sync plugins with gateway
2023-05-17T14:06:19Z DEBUG agent patch/manifest.go:97 found 3 plugins
2023-05-17T14:06:19Z DEBUG agent patch/manifest.go:153 loaded plugin manifest {"plugins": 3}
2023-05-17T14:06:19Z INFO agent v2/agent.go:516 received patch manifest from gateway
2023-05-17T14:06:19Z INFO agent patch/client.go:368 removing plugin: plugin_logging
2023-05-17T14:06:19Z INFO agent patch/client.go:368 removing plugin: plugin_metrics
2023-05-17T14:06:19Z INFO agent.pluginloader plugins/loader.go:150 loading plugin {"plugin": "github.com/rancher/opni/plugins/topology"}
2023-05-17T14:06:20Z INFO plugin.topology agent/plugin.go:52 node driver is available {"driver": "external-topology-operator"}
2023-05-17T14:06:20Z DEBUG agent.pluginloader plugins/loader.go:162 checking if plugin implements any interfaces in the scheme {"plugin": "github.com/rancher/opni/plugins/topology", "interfaces": ["opni.Health", "opni.apiextensions.HTTPAPIExtension", "opni.apiextensions.StreamAPIExtension", "opni.backends.CapabilityNode"]}
2023-05-17T14:06:20Z DEBUG agent.pluginloader plugins/loader.go:175 implementation found {"plugin": "github.com/rancher/opni/plugins/topology", "id": "opni.apiextensions.StreamAPIExtension"}
[...]

PR #1554, opened 2023-07-06
Fix up plugin loading

dbason:


dagger/main.go
```diff
@@ -404,6 +404,7 @@ func (b *Builder) runInTreeBuilds(ctx context.Context) error {
 		eg.Go(func() error {
 			charts.Directory(filepath.Join(b.workdir, "charts")).Export(ctx, "./charts")
 			charts.Directory(filepath.Join(b.workdir, "assets")).Export(ctx, "./assets")
+			charts.File(filepath.Join(b.workdir, "index.yaml")).Export(ctx, "./index.yaml")
```
kralicky, line 407: Are you sure this is what you want? `Export()` copies files back to the host itself, not between containers.

pkg/plugins/loader.go
```diff
@@ -278,7 +287,7 @@ func (p *PluginLoader) LoadPlugins(ctx context.Context, pluginDir string, scheme
 		md := md
 		clientOpts := options.clientOptions
 		if verifyManifest {
-			if secureConfig, ok := secureConfigs[md.Module]; ok {
+			if secureConfig, ok := secureConfigs[path.Base(md.Module)]; ok {
```
kralicky, line 290: Plugins are identified by the complete module path, just the base name isn't enough

dbason: That's not going to work in the new design because we don't have access to the full module path in the manifest; only the URN

alexandreLamarre (approved):
Fixes #1413 for me

alexandreLamarre (approved):

kralicky (approved):

PR #1561, opened 2023-07-11
Allow editing roles and role bindings

### LINKED ISSUE #1327: Allow editing roles
Currently, Role objects are immutable. We should create an edit API for them.

jan-law:
Fixes #1327


pkg/storage/stores.go
```diff
@@ -41,6 +41,7 @@ type ClusterStore interface {
 
 type RBACStore interface {
 	CreateRole(context.Context, *corev1.Role) error
+	UpdateRole(context.Context, *corev1.Role) error
```
kralicky, line 44: Can you add tests for this method to [RBACStoreTestSuite](https://github.com/rancher/opni/blob/24d21f0427ca7e41ea639fa18798f407f738e68c/pkg/test/conformance/storage/rbac_store.go#L15)?

kralicky (changes requested):

kralicky:
We also need to add the associated CLI commands to https://github.com/rancher/opni/blob/main/pkg/opni/commands/rbac.go

kralicky:
@codyrancher Do you have bandwidth to add UI for this? You can work directly in this branch.

codyrancher:
> @codyrancher Do you have bandwidth to add UI for this? You can work directly in this branch.

I'd prefer to do it in a separate PR so this code doesn't get held up.

pkg/opni/commands/rbac.go
```diff
@@ -169,6 +209,36 @@ func BuildRoleBindingsCreateCmd() *cobra.Command {
...
 }
 
+func BuildRoleBindingsUpdateCmd() *cobra.Command {
+	return &cobra.Command{
+		Use:   "update <rolebinding-id> <role-id> <user-id>...",
+		Short: "Update a role binding",
+		Args:  cobra.MinimumNArgs(3),
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			if len(args) == 1 {
+				return completeRoles(cmd, args, toComplete)
```
kralicky, line 219: This doesn't look quite right, can you update the completion logic for this command as well as the update roles command?

pkg/management/rbac.go
```diff
@@ -40,6 +61,28 @@ func (s *Server) CreateRoleBinding(ctx context.Context, in *corev1.RoleBinding)
...
+
+	oldRb, err := s.GetRoleBinding(ctx, in.Reference())
+	if err != nil {
+		return &emptypb.Empty{}, err
+	}
+
+	_, err = s.coreDataSource.StorageBackend().UpdateRoleBinding(ctx, oldRb.Reference(), func(rb *corev1.RoleBinding) {
+		rb.RoleId = in.GetRoleId()
+		rb.Subjects = in.GetSubjects()
+		rb.Taints = in.GetTaints()
```
kralicky, line 80: Can't store the taints field, it's read-only. Might also want to validate it in the implementation of UpdateRoleBinding as well to throw an error if it's non-nil

pkg/opni/commands/rbac.go
jan-law: Should https://github.com/rancher/opni/blob/main/pkg/opni/commands/rbac.go#L149 also be `completeRoleBindings` instead of `completeRoles`?

kralicky: It needs both, depending on the current number of arguments. Based on the usage 
`Use:   "create <rolebinding-id> <role-id> <user-id>` 
it should complete rolebindings for arg 0, roles for arg 1, and arg 2+ should just be ShellCompDirectiveNoFileComp

pkg/management/rbac.go
jan-law: working on validations now

pkg/storage/etcd/rbac_store.go
```diff
@@ -104,6 +220,7 @@ func (e *EtcdStore) GetRoleBinding(ctx context.Context, ref *corev1.Reference) (
 	if err := storage.ApplyRoleBindingTaints(ctx, e, roleBinding); err != nil {
```
jan-law, line 220: Question about role binding taints - why does the etcd and jetstream stores apply the taints in `GetRoleBinding` and not the crd store? 

My update code calls `GetRoleBinding` to perform an update with etcd or jetstream, and I end up with duplicate taints caused by `ApplyRoleBindingTaints` `rb.Taints = append(rb.Taints, "role not found")`.
image

 The crd rbac_store test doesn't even output taints if I get the role binding in a test.
```
Running Suite: CRDs Storage Suite - /home/ubuntu/opni/pkg/storage/crds
• [FAILED] [0.025 seconds]
RBAC Store Role Bindings [It] should handle taints for multiple role binding updates [integration, slow]
/home/ubuntu/opni/pkg/test/conformance/storage/rbac_store.go:219

  [FAILED] Expected
      <[]string | len:0, cap:0>: nil
  to equal
      <[]string | len:2, cap:2>: ["role not found", "no subjects"]
```
What's the expected behaviour?

kralicky: Role binding taints are purely an informative feature, the taints are computed on-demand and injected into the message before it's sent to the caller.

kralicky (approved):

Issue #1550, opened 2023-07-03
Generic CLI interactive editing


kralicky:
Currently, there is logic to support interactive editing of metrics capability configs using an editor (via `opni metrics config edit <id>`). This code should be made generic and integrated with other relevant commands. It should also be used by the CLI generator when generating commands that could utilize this functionality.

Issue #1549, opened 2023-07-03
Generic proto source generator for upstream cortex types


kralicky:
Implement a generic code generator that can generate protobuf message definitions from non-proto Go structs.
Intended for use on the upstream cortex config and alertmanager types, so that they can be used in plugin apis. It will also integrate with the CLI generator to automatically attach flag sets to cobra commands that are generated from RPC definitions.

OEP https://github.com/rancher/opni/pull/1254
Part of https://github.com/rancher/opni/issues/1264

Issue #1547, opened 2023-07-03
Generic Cortex configuration builder


kralicky:
The logic for generating the cortex config file in the test environment is different from the MonitoringCluster reconciler. We should replace these with a single generic config builder that can be used in both places, with the necessary features to allow each environment to use its own implementation-specific custom options.

OEP https://github.com/rancher/opni/pull/1254
Part of https://github.com/rancher/opni/issues/1264

Issue #1546, opened 2023-07-03
Cortex Runtime Config


kralicky:
Implement api types and management logic for the cortex runtime config.
OEP #1254
Part of #1264

PR #1563, opened 2023-07-11
Cortex Config refactor

kralicky:
Closes https://github.com/rancher/opni/issues/1550
Closes https://github.com/rancher/opni/issues/1549
Closes https://github.com/rancher/opni/issues/1547
Closes https://github.com/rancher/opni/issues/1546
Closes https://github.com/rancher/opni/issues/1357
Closes https://github.com/rancher/opni/issues/1264
Closes https://github.com/rancher/opni/issues/1601
Fixes https://github.com/rancher/opni/issues/1565
Fixes https://github.com/rancher/opni/issues/1015
Fixes https://github.com/rancher/opni/issues/712
Fixes https://github.com/rancher/opni/issues/807

Part of epic https://github.com/rancher/opni/issues/1262

Lots of things moved/renamed in here, but the important changes are concentrated in a few places, in order of importance:
1. `pkg/plugins/driverutil` - contains new generic api-based configuration management tools for plugin drivers
2. `pkg/plugins/apis/system/` - contains enhancements to the kv store interfaces
3. `pkg/storage/crds/value_store.go` - contains a new generic value store backed by crd objects (also used for plugin drivers)
	- there are many other updates in pkg/storage/ to implement new features and updated interfaces, but nothing interesting)
4. `plugins/metrics/apis/cortexops/`, `plugins/metrics/pkg/cortex/configutil/` - contains the new cortexops api implementation. lots of other small changes in plugins/metrics but most of the new code is in these two packages.
	- also some controller updates in `controllers/core_monitoring_controller.go` which handles automatically upgrading existing MonitoringClusters as to not break existing installs
6. `internal/codegen/` for enhancements to the cli generator and descriptor generator
7. `pkg/opni/cliutil/`, `pkg/util/flagutil/`, `pkg/tui/` for some new (+ some moved) cli tools

To get a better overview of the relevant changes, `git diff --stat --minimal origin/main -- . ':^*.yaml' ':^*.pb*.go' ':^*zz_generated*' ':^*_test.go' ':^pkg/test/mock'`


dbason (approved):

dbason (dismissed):

pkg/resources/monitoring/grafana.go
```diff
@@ -175,9 +175,14 @@ func (r *Reconciler) grafana() ([]resources.Resource, error) {
 		return nil, fmt.Errorf("invalid grafana hostname: %w", err)
 	}
 
+	tag := "latest"
```
alexandreLamarre, line 178: do we really want to pin this to `latest` ? I'm assuming any major changes to datasource apis will be matched in the latest cortex version, so I'm also assuming this is ok

pkg/resources/monitoring/image.go
```diff
@@ -2,23 +2,16 @@ package monitoring
...
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 func (r *Reconciler) updateImageStatus() (bool, error) {
 	lg := r.logger
-	var image string
-	var pullPolicy corev1.PullPolicy
-	if imgOverride := r.mc.Spec.Cortex.Image.GetImageWithDefault(""); imgOverride != "" {
```
alexandreLamarre, line 14: It might still make sense to allow overriding the image, we could always open an issue to reintroduce this later

pkg/test/environment.go
```diff
@@ -1383,6 +1456,11 @@ func (e *Environment) NewGatewayConfig() *v1beta1.GatewayConfig {
 				GRPCListenAddress: fmt.Sprintf("tcp://localhost:%d", e.ports.ManagementGRPC),
 				HTTPListenAddress: fmt.Sprintf("localhost:%d", e.ports.ManagementHTTP),
 				WebListenAddress:  fmt.Sprintf("localhost:%d", e.ports.ManagementWeb),
+				// WebCerts: v1beta1.CertsSpec{
+				// 	CACertData:      dashboardCertData,
+				// 	ServingCertData: dashboardCertData,
+				// 	ServingKeyData:  dashboardKeyData,
+				// },
```
alexandreLamarre, line 1463: keep or remove?

pkg/test/testdata/testdata/cortex/config.yaml
```diff
@@ -1,115 +0,0 @@
...
-  grpc_client_config: *tls_client_config
-  match_max_concurrent: true
-
-ingester_client:
-  grpc_client_config: *tls_client_config
-
-ruler:
-  evaluation_interval: 1s
-  poll_interval: 1s
-  resend_delay: 1s
```
alexandreLamarre, line 45: these values are important to keep as default for the test environment, so wherever these are set now we need to make sure they are the same

alexandreLamarre (changes requested):
Small questions mostly

controllers/suite_test.go
```diff
@@ -88,11 +88,11 @@ const (
 )
 
 func TestAPIs(t *testing.T) {
-	SetDefaultEventuallyTimeout(30 * time.Second)
+	SetDefaultEventuallyTimeout(10 * time.Second)
```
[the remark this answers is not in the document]
kralicky: Good catch yeah

magefiles/targets/test.go
```diff
@@ -27,6 +27,16 @@ func (Test) All() error {
 	}, mg.GoCmd(), "test", "-race", "./...")
 }
 
+// Runs all tests in short mode
+func (Test) Short() error {
+	if testbinNeedsUpdate() {
+		mg.Deps(Test.Bin)
+	}
+	return sh.RunWithV(map[string]string{
+		"CGO_ENABLED": "1",
```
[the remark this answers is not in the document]
kralicky: CGO_ENABLED is explicitly set to 0 in CI, but the race detector requires cgo

pkg/dashboard/dashboard.go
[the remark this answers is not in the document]
kralicky: UI changes are not complete yet.

pkg/resources/monitoring/grafana.go
kralicky: Older versions didn't have the ability to set the tag, so if it's not set in the config the default is `latest` to match the old behavior. It can (and probably should) be pinned to a specific version, but that should be configured elsewhere. `latest` is semantically equivalent to empty tag so it's a reasonable default.

pkg/resources/monitoring/image.go
kralicky: Yeah, the grafana config here is very much still outdated and needs to be addressed. That's a whole other thing because we have to figure out what to do with the upstream grafana operator. I tried to make as few changes to it as I could in this pr.

pkg/test/environment.go
kralicky: Keep, I want to be reminded about this later on. lol

pkg/test/testdata/testdata/cortex/config.yaml
kralicky: They are set here https://github.com/rancher/opni/blob/ff9ffbdb415585a1debe0b3268a358147c32f3ae/plugins/metrics/test/test_drivers.go#L271-L276

kralicky: Looks like it's missing ResendDelay though so I'll add that

dbason (dismissed):

pkg/dashboard/dashboard.go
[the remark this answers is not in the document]
alexandreLamarre: @codyrancher @kralicky UI needs to be potentially fixed and tested in a kubernetes cluster, see repro steps below:

image
image
Then selecting HA preset breaks the UI : 
image
image

alexandreLamarre (dismissed):

alexandreLamarre (commented):
Actually still seeing some UI crashes with the current build

dbason (approved):

PR #1569, opened 2023-07-17
Add support label and role creation

### LINKED ISSUE #1568: Opni does not reconcile multiclusteruser objects
The reconciler is not loaded in the opni manager

dbason:
This PR does the following things:

- Adds a well known Opni label to designate a cluster is actually a support user
- Creates a user account for Opensearch when the logging capability is added to a support user
- Generates the password for that account ratcheted off the client key
- Adds a command to the support CLI to display the generated password.

Fixes #1568

### DIFF
--- a/pkg/crypto/xof.go
+++ b/pkg/crypto/xof.go
@@ -1,0 +1,37 @@
+package crypto
+
+import (
+	"errors"
+
+	"golang.org/x/crypto/sha3"
+)
+
+type PRFXOFHasher interface {
+	Hash(data []byte, outputLen int) ([]byte, error)
+}
+
+type cshakeHasher struct {
+	domain []byte
+	key    []byte
+}
+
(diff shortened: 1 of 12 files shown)


kralicky (dismissed):

pkg/crypto/xof.go
```diff
@@ -0,0 +1,37 @@
...
+type PRFXOFHasher interface {
+	Hash(data []byte, outputLen int) ([]byte, error)
+}
+
+type cshakeHasher struct {
+	domain []byte
+	key    []byte
+}
+
+func NewCShakeHasher(key, domain []byte) PRFXOFHasher {
```
kralicky, line 18: suggest naming this variable "secret" or "secretKey" to disambiguate. Also, domain can be a string to prevent accidentally reversing the order of arguments here.

kralicky (approved):

Issue #1573, opened 2023-07-17
Minimal image environment variable contents discrepency


kralicky:
The `OPNI_MINIMAL_IMAGE_REF` environment variable set at https://github.com/rancher/opni/blob/main/pkg/oci/kubernetes/kubernetes.go#L96-L109 is being parsed by code at https://github.com/rancher/opni/blob/main/pkg/oci/kubernetes/kubernetes.go#L96-L109 in the wrong format. It assumes the environment variable only contains the digest (e.g. `"sha256:..."`) but it actually contains the entire image reference (e.g. `"docker.io/rancher/opni:latest-minimal@sha256:..."`)

PR #1576, opened 2023-07-17
Fix minimal image ref format inconsistencies in image update logic

kralicky:
Fixes https://github.com/rancher/opni/issues/1573

### DIFF
--- a/pkg/oci/kubernetes/images.go
+++ b/pkg/oci/kubernetes/images.go
@@ -7,16 +7,14 @@ import (
 	"time"
 
 	corev1beta1 "github.com/rancher/opni/apis/core/v1beta1"
-	"github.com/rancher/opni/pkg/versions"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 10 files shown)


pkg/oci/kubernetes/images.go
```diff
@@ -40,34 +39,26 @@ func (d *kubernetesResolveImageDriver) getOpniImageString(ctx context.Context) s
...
-		return envDigest
+func getMinimalImageString() (string, error) {
+	envRef := os.Getenv(minimalImageRefEnv)
+	if envRef != "" {
+		return envRef, nil
 	}
 
 	if versions.Version != "unversioned" {
-		return fmt.Sprintf("%s-minimal", versions.Version)
+		return fmt.Sprintf("%s-minimal", versions.Version), nil
```
dbason, line 60: This should probably be something like 
```go
return fmt.Sprintf("rancher/opni:%s-minimal", versions.Version), nil
```

dbason (changes requested):

dbason (approved):

Issue #1529, opened 2023-06-22
Alerting : Alarm Filter Options


alexandreLamarre:
The following proto message

https://github.com/rancher/opni/blob/a7bdc9152a67f7661721cee0db58bd394475b8b0/pkg/apis/alerting/v1/alerting.condition.proto#L433-L438|

can be used in requests to the apis for :
- ListAlarms
- BatchStatus
- Timeline
- Group ids

codyrancher:
Some questions for this one.

- For ListAlertConditionRequest what's the logic operations used for the properties? Is it `AND` for each property i.e. this will only show entities that have both the listed cluster and the listed severity? Are the lists `OR` i.e. I want it for this cluster or this cluster?
- Are you wanting to specify more than one filter at a time?
- Are the other backends going to support similar filter types?

alexandreLamarre:
this didn't pop into my notifications at the time.
The filters are AND -ed togther,  a single filter object can be specfied at a time.

> Are the other backends going to support similar filter types? 

 Only plan is to support conditions with these filters, but they apply to other APIs where we list conditions : status, timeline, etc...

PR #1579, opened 2023-07-19
Adding group-id to alarms

codyrancher:
Adding group-id to alarms

https://github.com/rancher/opni/assets/55104481/c935c789-3e5d-4e36-8b3a-708a11718436

The test-env doesn't seem to return anything but the `''` group id when set. If I'm missing something let me know and I'll fix it.


alexandreLamarre (changes requested):
The default group is `""`, so we should show that as `Default` .

There's a bug in the backend currently when updating groups (groups will only be respected when being created). The list/batch API must include which groups are being listed, so I have to change the list/batch status APIs to post requests so you can post the filters/groupIds to list. That would also mean that the group filters would have to be populated by `/AlertConditions/groups` API.

codyrancher:
Updated to use ConditionReference everywhere and updated the group `''` to be Default instead of None.

image
image

alexandreLamarre (changes requested):
It looks like the UI lists all groups by default -- only the default group should be shown :  the main motivation for having the grouping mechanism is that the sync mechanism for prometheus rules will sync downstream rules and bucket them.

To clarify the expected size of buckets and number of alarms used:

- Each metric exporter libraries bundle their own rules, usually several dozen, which means per cluster we will have the order of 100-200 alarms
- In a multi cluster setting this quickly balloons to 1000s of alarms
- Each synced group will likely have 10-20 alarms each.
- Synced groups may overlap between clusters but also may not, depending on which metric exporters are installed.


Therefore we should replace the text filter with a text filter that filters a drop down of available groups. Since we don't necessarily create these groups ourselves, we should limit the dropdown to (10?) and have the text filter responsible for filtering the groups the user wishes select

Similarly we would really need  https://github.com/rancher/opni/issues/1529 to limit the sheer amount of alarms displayed and have some sane defaults like only displaying the local cluster by default

codyrancher:
Here's an updated video
https://github.com/rancher/opni/assets/55104481/b0ea1d0a-f2ef-488f-98bb-041a22ecda54

I don't think we need to have a max group count in the dropdown since it's possible to search and select. I'll work on adding the other filters in https://github.com/rancher/opni/issues/1529. I may talk to someone else about other options because I think adding so many filters is going to clutter the view but I'll cross that road once I start on that one.

alexandreLamarre (approved):

PR #1588, opened 2023-07-20
WIP: Parrallel import

joshmeranda:
Support parallel read and write with internal data buffer

### DIFF
--- a/plugins/metrics/pkg/agent/runner.go
+++ b/plugins/metrics/pkg/agent/runner.go
@@ -153,13 +153,6 @@ func (tr *taskRunner) SetRemoteWriteClient(client clients.Locker[remotewrite.Rem
 	tr.remoteWriteClient = client
 }
 
-// func (tr *taskRunner) SetRemoteReaderClient(client RemoteReader) {
-// 	tr.remoteReaderMu.Lock()
-// 	defer tr.remoteReaderMu.Unlock()
-
-// 	tr.remoteReader = client
-// }
-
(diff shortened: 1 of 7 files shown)


plugins/metrics/pkg/agent/runner.go
```diff
@@ -203,96 +195,157 @@ func (tr *taskRunner) doPush(ctx context.Context, writeRequest *prompb.WriteRequ
...
 
-			writeRequest := prompb.WriteRequest{
-				Timeseries: dereferenceResultTimeseries(result.Timeseries),
+			activeTask.AddLogEntry(zapcore.InfoLevel, fmt.Sprintf("split request into %d chunks", len(chunks)))
+
+			lo.ForEach(chunks, func(chunk *prompb.WriteRequest, i int) {
+				if err := b.Add(wc, WriteMetadata{
+					Query:         readRequest.Queries[0],
+					WriteChunk:    chunk,
+					ProgressRatio: 1.0 / float64(len(chunks)),
```
jan-law, line 298: Do all chunks have the same progress ratio?

joshmeranda: All chunks from the same original `prompb.WriteRequest` do, but will differ between chunks from different `prompb.WriteRequest`.

Not super happy with the way I currently have this implemented. Will probably change this to a single progress delta so I can get rid of the `Query` field and the `ProgressRatio` all the same logic will remain essentially the same just moved to before data is added to the buffer, but we'll be storing less data.

PR #1590, opened 2023-07-20
send alerts to k8s events

### LINKED ISSUE #1544: Add kubernetes events endpoint for alerts
Should provide an alerting endpoint that writes alerts to kubernetes as events.

alexandreLamarre:
Configures the Alertmanager instance to publish opni alerts to kubernetes events

Closes #1544


dbason (commented):
As discussed we should take a look at https://pkg.go.dev/k8s.io/client-go/tools/record to see if we can reuse that.

dbason (approved):

PR #1606, opened 2023-08-02
Fix various metrics test flake issues

kralicky:


pkg/test/environment.go
```diff
@@ -463,6 +463,9 @@ func (e *Environment) StartEmbeddedJetstream() (*nats.Conn, error) {
 }
 
 func (e *Environment) Stop() error {
+	os.Unsetenv("NATS_SERVER_URL")
```
kralicky, line 466: btw this environment variable is the cause of nats server-related test flakes when multiple test environments are in use at the same time. @alexandreLamarre

alexandreLamarre (approved):

dbason (approved):

dbason (approved):

PR #1607, opened 2023-08-02
logging: create Trace pipeline from OTel to OpenSearch

### LINKED ISSUE #1605: Opni logging should collect traces
This will involve:

1. Implementing Export methods for traces in the otel forwarder
2. Adding a trace pipeline to the otel preprocessor config
3. Implementing an Export method for traces in the gateway
4. Adding a trace pipline to the otel config for logging.

jaehnri:
Closes #1605 

This PR implements a pipeline for traces to flow the same way logs do:
```
Aggregator -> Agent Forwarder -> Preprocessor -> OpenSearch
```

## OpenSearch

A new index called `ss4o_traces-kubernetes-opni` was created using the `opni-indexer` permissions. Using the OpenSearch DevTools, it is possible to fetch spans from the index:
image

## Testing Environment

These changes were tested using a newly deployed K3s cluster (v1.26.8 +k3s1).
To create a constant mass of traces for testing, the [Kubernetes System Traces](https://kubernetes.io/docs/concepts/cluster-administration/system-traces/) Alpha feature was used with the following configuration:

/etc/rancher/k3s/trace-config.yaml:
```yaml
apiVersion: apiserver.config.k8s.io/v1alpha1
kind: TracingConfiguration
# Aggregator ClusterIP
endpoint: 10.43.111.6:4317
samplingRatePerMillion: 10000
```
/etc/systemd/system/k3s.service:
```
...
ExecStart=/usr/local/bin/k3s \
    server \
    '--kube-apiserver-arg=feature-gates=APIServerTracing=true' \
    '--kube-apiserver-arg=tracing-config-file=/etc/rancher/k3s/trace-config.yaml' \
```

An integration test was also added to simulate the same flow in `forwarder_test.go`. It deploys two embedded OTel Collectors, a gRPC server acting like the Forwarder and a Sink server. It ensures that the pipeline works by sending a Trace to the aggregator collector and checking that it reaches the final destination (sink server).

## Known Issues

For some reason, the new index won't show in the Discover tab, used for overall navigation. The remaining OpenSearch features work fine.


pkg/resources/collector/workloads.go
```diff
@@ -46,13 +46,17 @@ func (r *Reconciler) aggregatorConfigMapName() string {
 	return fmt.Sprintf("%s-aggregator-config", r.collector.Name)
 }
 
-func (r *Reconciler) getDaemonConfig(loggingReceivers []string) otel.NodeConfig {
+func (r *Reconciler) getDaemonConfig(loggingReceivers []string, traceReceivers []string) otel.NodeConfig {
```
jaehnri, line 49: The trace receivers are push-based, so I don't think they should be configured as daemons, but I don't see how else we could do. Any thoughts?

pkg/resources/collector/templates.go
```diff
@@ -138,6 +139,14 @@ filelog/rke2:
     value: kubelet
 `))
 
+	templateTraceReceiver = `
```
jaehnri, line 142: This configuration is static, so I'm not sure if it should be a template. I'm also unsure about how these templates are actually used in the code.

pkg/resources/collector/workloads.go
alexandreLamarre: The way I see it there are two options for the daemonset:
 1. Not using the daemonset at all: 
 So something along the lines of : 
```go
func (r *Reconciler) requiresDaemoset() {
  return r.Config.Logging != nil || r.Config.Metricss != nil
}
```

And in the method where the daemonset is reconciled, something along the lines of : 
```
if !r.requiresDaemonset() {
  resources.Absent(daemonset)
} else {
   // ...
   resources.Present()
}
```
2. (better) collecting the internal traces from the open-telemetry collector:

The open telemetry collector collects self-telemetry from the following service:
```yaml
service:
  telemetry:
      logs:
        level: {{ .LogLevel }}
      metrics:
        level: none
      traces:
        // TODO
```

Note : you should definitely include the self-telemetry for at least the aggregator

pkg/resources/collector/logging.go
```diff
@@ -184,3 +184,26 @@ func (r *Reconciler) hostLoggingVolumes() (
...
+	})
+	retVolumeMounts = append(retVolumeMounts, corev1.VolumeMount{
+		Name:             "otel-traces",
+		ReadOnly:         true,
+		MountPath:        otelConfigDir,
+		MountPropagation: lo.ToPtr(corev1.MountPropagationHostToContainer),
+	})
+
+	return
+}
```
alexandreLamarre, line 209: I could be wrong, but I don't think we need any additional volumes for traces, traces don't require any additional persistent storage. 

I see here that you're attempting to mount a config,  the config for this collector is already in the template written to the config map.  Each of metrics, logs and traces use the same config.

[daemon config map](https://github.com/rancher/opni/blob/98ac856f97a4ecabf8ac5eff31c89d6c5a48c46d/pkg/resources/collector/workloads.go#L113)

[aggregator config map](https://github.com/rancher/opni/blob/98ac856f97a4ecabf8ac5eff31c89d6c5a48c46d/pkg/resources/collector/workloads.go#L152)

[volumes mounting the daemon set config map](https://github.com/rancher/opni/blob/98ac856f97a4ecabf8ac5eff31c89d6c5a48c46d/pkg/resources/collector/workloads.go#L192-L209)

[volumes mounting the aggregator config map](https://github.com/rancher/opni/blob/98ac856f97a4ecabf8ac5eff31c89d6c5a48c46d/pkg/resources/collector/workloads.go#L327-L345)

pkg/resources/collector/templates.go
alexandreLamarre: You are correct you don't need a template here, since below you are already embedding the trace config into the template which gets translated to the yaml config the collector uses

pkg/otel/types.go
```diff
@@ -42,6 +44,11 @@ type LoggingConfig struct {
 	Receivers []string
 }
 
+type TraceConfig struct {
+	Enabled   bool
+	Receivers []string
```
alexandreLamarre, line 49: Probably don't need an array of receivers in the `TraceConfig` since these will be static for now

plugins/logging/pkg/gateway/stream.go
```diff
@@ -16,8 +17,13 @@ func (p *Plugin) StreamServers() []streamext.Server {
...
 		{
 			Desc:              &collogspb.LogsService_ServiceDesc,
-			Impl:              p.otelForwarder,
+			Impl:              p.otelForwarder.LogsForwarder,
 			RequireCapability: wellknown.CapabilityLogs,
 		},
+		{
+			Desc:              &coltracepb.TraceService_ServiceDesc,
+			Impl:              p.otelForwarder.TraceForwarder,
+			RequireCapability: wellknown.CapabilityTraces,
```
alexandreLamarre, line 26: this should be `wellknown.CapabilityLogs`, unless we really want to introduce logic inside the logging agent capability api to toggle logs/traces separately.

@dbason I think this is what you wanted for the initial pass to be for traces?

plugins/logging/pkg/otel/trace.go
```diff
@@ -0,0 +1,166 @@
...
+	if len(values)%2 != 0 {
+		f.lg.Warnf("invalid number of attribute values: %d", len(values))
+		return f.forwardTrace(ctx, request)
+	}
+
+	for i := 0; i < len(values); i += 2 {
+		key := values[i]
+		value := values[i+1]
+		addTraceToResourceSpans(request, key, value)
+	}
```
alexandreLamarre, line 114: @dbason do we need support for traces for the support agent?

pkg/resources/collector/templates.go
alexandreLamarre: @jaehnri you will have to also [implement this for traces within this method](https://github.com/rancher/opni/blob/98ac856f97a4ecabf8ac5eff31c89d6c5a48c46d/plugins/logging/pkg/agent/drivers/kubernetes_manager/kubernetes_manager.go#L122-L139)

pkg/resources/collector/logging.go
jaehnri: Makes sense. For some reason, my initial understanding was that the collector worked like a database (?). For that reason, I thought we needed a separate volume. Thanks!

plugins/logging/pkg/gateway/stream.go
kralicky: Technically this field doesn't do anything yet, but it's reserved for future use, but yeah it should be `wellknown.CapabilityLogs`. If we want traces to be separate then we would make a separate plugin for it and then it would be its own capability. Inside the logging plugin it's all part of the logging capability.

pkg/resources/preprocessor/workloads.go
```diff
@@ -70,31 +72,39 @@ processors:
...
+      dedup: true
+      dedot: false
 service:
   pipelines:
     logs:
       receivers: ["otlp"]
       processors: ["resource", "attributes", "transform"]
       exporters: ["opensearch"]
+    traces:
+      receivers: ["otlp"]
```
jaehnri, line 106: I ended up not adding any processors for the trace pipelines. Will do them in https://github.com/rancher/opni/issues/1748.

pkg/resources/preprocessor/workloads.go
```diff
@@ -70,31 +72,39 @@ processors:
...
-    endpoints: [ "{{ .Endpoint }}" ]
-    index: {{ .WriteIndex }}
+    http:
+      endpoint: {{ .Endpoint }}
+      tls:
+        ca_file: /etc/otel/chain.crt
+        cert_file: /etc/otel/certs/tls.crt
+        key_file: /etc/otel/certs/tls.key
+    logs_index: {{ .WriteIndex }}
+    dataset: kubernetes
```
jaehnri, line 91: Right now, the OpenSearch trace exporter only supports indexing according to the Observability plugin [index naming conventions](https://opensearch.org/docs/latest/observing-your-data/ss4o/#traces-and-metrics), i.e., `ss4o_traces-dataset-namespace`. 

For that reason and to avoid confusion for end users, I hardcoded this index name here. 

However, we are using our own fork, so we could change those index naming conventions to use any index name that we like and allow our users to do so, as logs are done. What do you think?

pkg/resources/preprocessor/workloads.go
```diff
@@ -20,10 +20,12 @@ import (
...
-	configKey             = "config.yaml"
-	preprocessorVersion   = "v0.1.4-rc1-0.85.0"
-	preprocessorImageRepo = "ghcr.io/rancher-sandbox"
-	preprocessorImage     = "opni-otel-collector"
+	configKey = "config.yaml"
+
+	// TODO: Change back to opni-otel-collector image
+	preprocessorVersion   = "opensearch-exporter"
+	preprocessorImageRepo = "docker.io/jaehnri"
+	preprocessorImage     = "otel-collector"
```
alexandreLamarre, line 28: We should switch to a release of rancher-sandbox, unless something prevents it

jaehnri: Yup, it is waiting for https://github.com/rancher-sandbox/opni-otel-collector/pull/12

alexandreLamarre (dismissed):

dbason (approved):

alexandreLamarre (approved):

PR #1613, opened 2023-08-03
Update versions for new RC

dbason:


ron1:
@dbason Are you able to publish updated 0.11 rc  helm charts as well?

Issue #1170, opened 2023-03-11
System-wide logging improvements


kralicky:
The zap logger setup we have has been working well so far, but as the project expands we should consider:
a) potential improvements, refactors, or redesigns of how we use loggers across places in the code
b) newer/better/standardized libraries to switch to

I've tried a few other logging libraries, and so far none seem to be as good as zap for our use cases, but more research is required.

### slog (upcoming standard library structured logging)
- It has a design similar to zap, and a good feature set and API
- It is still very much in development, and lacks a lot of features that I like from zap (e.g. custom log output handlers)

### zerolog
- Inspired by/derived from zap, but with fewer options for customization. For this reason it probably isn't a good fit.
- Performance improvements are very interesting though - I've noticed that zap is pretty heavy on memory allocations (although there are a lot of optimizations we can make in our own code).

-------

#### Some issues we are running into currently:
- Plugin logs are weird. I'm considering making a fork of go-plugin and swapping out hclog because the interoperability with zap is complicated
- Sampling lacks important features such as logging the number of dropped logs. I've implemented this in my fork of zap which we use, but it's annoying to maintain.

jan-law:
As discussed earlier, noting some additional improvement suggestions
- fetch agent plugin logs remotely instead of manually
- send test log output to a file
- I've kept the sampling behaviour the same as before. However, slog-sampling lets you configure the % rate of sampled logs for each log level (eg log 10% Info, 100% error and above). Suggestions welcome

PR #1642, opened 2023-08-10
Refactor loggers with Slog

jan-law:
Related #1170
Depends on https://github.com/dbason/opentelemetry-collector-contrib/pull/1
Depends on https://github.com/kralicky/totem/pull/5

Replaces zap with slog. According to docs, this should eliminate any allocations for logging strings and numbers. Benchmark test results for slog handlers: https://gist.github.com/jan-law/bfb4db82ba05dbcb551e0de1ed36f8c1

### How to rebase branches ahead of main
In general, logging calls are identical with a few exceptions:
- Format string logs eg `Infof`, `Errorf` -> replace with `Info(fmt.Sprintf(...))`, `Error(fmt.Sprintf(...))`
- `logger.New().Named("some-name")` -> replace with `logger.New().WithGroup("some-name")`
- When you need to create a `logr.Logr`, you can use `logger.NewLogr(options...)`
- To create a new logger, see `pkg/logger/logger.go` for available options

[This script](https://gist.github.com/jan-law/45fc8e42983b63a64eed4d8bb920a5e2) will help you refactor existing zap logs with the slog equivalents.

Disclaimer: gofmt can only refactor logging calls that have a fixed number of arguments and is also case sensitive. If you created a log message with a different number of key/val pairs than what already exists in the codebase, the script may not format that log message.
 
### Notes
- log messages of the form `logger.With(args...).Info("blah")` are still valid, but may cost extra memory allocations. You can replace them with `logger.Info("blah", args...)`
- Slog doesn't enforce an even number of args when you're logging loosely typed key/value pairs. It will log the last odd-numbered arg as `!BADKEY=<value>`. Docs recommend using https://pkg.go.dev/cmd/vet to check your code


jan-law:
Note: I still need to update go.mod with the updated commit hash once https://github.com/dbason/opentelemetry-collector-contrib/pull/1 is merged

kralicky:
Some of the log messages that look like this:
```go
a.logger.With(
	zap.Error(err),
	zap.String("group", ruleGroup.Name),
).Error("failed to marshal rule group")
```
seem to have been rewritten as
```go
a.logger.Error("failed to marshal rule group", logger.Err(err),
	"group", ruleGroup.Name)
```
Is there a way we can have the format match? i.e.
```go
a.logger.With(
	logger.Err(err),
	"group", ruleGroup.Name,
).Error("failed to marshal rule group")
```

jan-law:
> Is there a way we can have the format match? 

Yes, your last example is valid slog syntax. The reason why I changed the format is because the original format causes an extra allocation per line, and you mentioned you wanted to reduce allocations by moving to slog.

I can revert the format of the `logger.With` calls so that they match if you'd prefer the same format?

kralicky:
Ah, yeah I think for most cases the extra memory allocation isn't a big deal, but in performance oriented code that is good to keep in mind. we should keep the old format in places where it's originally written like that.

kralicky (dismissed):
LGTM

jan-law:
Rebased onto main, no other changes

jan-law:
@kralicky This PR is using my fork of Dan's copy of `opentelemetry-collector-contrib`. Dan mentioned his copy is outdated and needed time to think about how to update the rancher sandbox fork [here](https://github.com/rancher-sandbox/otel-collector-contrib). 

Should we merge this PR and then update the fork later? How do you want to handle this?

kralicky:
Yeah that sounds fine to me, we can merge this PR first.

kralicky (approved):

PR #1666, opened 2023-08-18
Collect audit logs from kube-apiserver

### LINKED ISSUE #1574: Opni should collect audit logs
Today rancher-logging collects both Rancher and Kubernetes audit logs.  Opni should also do this.

jaehnri:
This is how the audit logs are shown in OpenSearch:
```json
{
    "_id": "5f2b9baf-cbd4-44bf-b81f-b5ad28e0aff1",
    "_index": "logs-v0.5.4-000003",
    "_score": null,
    "_source": {
        "anomaly_level": "",
        "Body.value": "{\"kind\":\"Event\",\"apiVersion\":\"audit.k8s.io/v1\",\"level\":\"Metadata\",\"auditID\":\"234fa224-b400-404e-931e-46effd64a321\",\"stage\":\"ResponseStarted\",\"requestURI\":\"/api/v1/pods?allowWatchBookmarks=true\\u0026resourceVersion=3963946\\u0026timeout=5m12s\\u0026timeoutSeconds=312\\u0026watch=true\",\"verb\":\"watch\",\"user\":{\"username\":\"system:kube-controller-manager\",\"groups\":[\"system:authenticated\"]},\"sourceIPs\":[\"127.0.0.1\"],\"userAgent\":\"k3s/v1.26.7+k3s1 (linux/amd64) kubernetes/e47cfc0/shared-informers\",\"objectRef\":{\"resource\":\"pods\",\"apiVersion\":\"v1\"},\"responseStatus\":{\"metadata\":{},\"code\":200},\"requestReceivedTimestamp\":\"2023-08-30T19:46:38.513630Z\",\"stageTimestamp\":\"2023-08-30T19:46:38.514778Z\",\"annotations\":{\"authorization.k8s.io/decision\":\"allow\",\"authorization.k8s.io/reason\":\"RBAC: allowed by ClusterRoleBinding \\\"system:kube-controller-manager\\\" of ClusterRole \\\"system:kube-controller-manager\\\" to User \\\"system:kube-controller-manager\\\"\"}}",
        "cluster_id": "e1967963-f4ec-4e4a-afa1-a766ff0ddb18",
        "deployment": "",
        "k8s.auditlog.audit_id": "234fa224-b400-404e-931e-46effd64a321",
        "k8s.auditlog.level": "Metadata",
        "k8s.auditlog.resource": "pods",
        "k8s.auditlog.stage": "ResponseStarted",
        "k8s.auditlog.stage_timestamp": "2023-08-30T19:46:38.514778Z",
        "kubernetes_component": "",
        "log": "",
        "log_type": "controlplane",
        "namespace_name": "",
        "pod_name": "",
        "service": "",
        "SeverityNumber": 0,
        "template_matched": "",
        "time": 1693424798514,
        "TraceFlags": 0
    },
    "_version": 1,
    "fields": {
        "time": [
            "2023-08-30T19:46:38.514Z"
        ]
    },
    "sort": [
        1693424798514
    ]
}
```

As the audit log is a bit complex, I chose a few properties that make more sense to be parsed from the JSON log:
- "k8s.auditlog.audit_id": "234fa224-b400-404e-931e-46effd64a321",
- "k8s.auditlog.level": "Metadata";
- "k8s.auditlog.resource": "pods";
- "k8s.auditlog.stage": "ResponseStarted";
- "k8s.auditlog.stage_timestamp": "2023-08-30T19:46:38.514778Z"

The whole log can still be checked in the body.
image

## Testing config

For testing, I used the following Kubernetes auditLog configuration:
```yaml
apiVersion: audit.k8s.io/v1
kind: Policy
rules:
  - level: Metadata
    resources:
    - group: ""
      resources: ["pods"]
```

And this logging.opni.io configuration in a K3s cluster:
```yaml
...
spec:
  kubeAuditLogs:
    auditFilename: audit.log
    enabled: true
    pathPrefix: /var/lib/rancher/k3s/server/logs
  provider: k3s
  selector: {}
```


pkg/resources/collector/workloads.go
```diff
@@ -559,3 +578,13 @@ func (r *Reconciler) configReloaderImageSpec() opnimeta.ImageSpec {
 		}(),
 	}.Resolve()
 }
+
+func (r *Reconciler) fetchCollectorConfig() (*opniloggingv1beta1.CollectorConfig, error) {
```
alexandreLamarre, line 582: nit : rename to `fetchLoggingCollectorConfig`

pkg/resources/collector/templates.go
```diff
@@ -101,6 +102,49 @@ journald/k3s:
   units: [ "k3s" ]
   directory: {{ . }}
 `))
+
+	templateKubeAuditLogs = `
+filelog/kubeauditlogs:
+  include: [ /var/log/kube-audit/*.log ]
```
alexandreLamarre, line 108: We should probably be aware that this hard-coded value may not always be correct https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/#log-backend. 

I doubt there is a way to programmatically discover which path is in use by the apiserver, but figure I'd ask anyways

We should check (if you haven't already) that this prefix is the default for all the Kubernetes distributions we currently support with logs, EKS, RKE1/2, K3s

jaehnri: Agreed. From what I've seen in the docs, there's no "default" audit log location and it does not pair that well across the distributions.

I've seen [this small "how to" from SUSE](https://www.suse.com/support/kb/doc/?id=000021022) showing how one can activate audit logs with rancher-logging. In this case, if no LogPath is provided in the CollectorConfigCRD, it uses the article's path. 

I confess I'm still confused about how the reconciling process works for the logging plugin, so I'm not sure if the CRD is really reconciled and, thus, if the LogPath does get picked up somewhere.

pkg/resources/collector/templates.go
```diff
@@ -101,6 +102,53 @@ journald/k3s:
...
+    to: resource["k8s.auditlog.stage_timestamp"]
+  - type: move
+    from: attributes.level
+    to: resource["k8s.auditlog.level"]
+  - type: move
+    from: attributes.auditID
+    to: resource["k8s.auditlog.audit_id"]
+  - type: move
+    from: attributes.objectRef.resource
+    to: resource["k8s.auditlog.resource"]
```
jaehnri, line 138: 1) Should these 5 fields be indexed?
2) Where do we select which fields are indexed?

pkg/resources/collector/templates.go
```diff
@@ -101,6 +102,53 @@ journald/k3s:
...
+  include_file_name: false
+  operators:
+  - type: json_parser
+    id: parse-body
+    timestamp:
+      parse_from: attributes.stageTimestamp
+      layout: '%Y-%m-%dT%H:%M:%S.%LZ'
+  - type: add
+    field: attributes.log_type
+    value: controlplane
```
dbason, line 120: I don't think we should tag these as controlplane as that has implications for the controlplane ai models

pkg/resources/collector/templates.go
dbason: I think leaving these is fine.  Any fields that are included will get indexed

dbason (changes requested):

apis/logging/v1beta1/collector_config_types.go
```diff
@@ -49,6 +49,11 @@ type RKE2Spec struct {
 	LogPath string `json:"logPath,omitempty"`
 }
 
+type KubeAuditLogsSpec struct {
+	Enabled bool   `json:"enabled,omitempty"`
+	LogPath string `json:"logPath,omitempty"`
```
jaehnri, line 54: @alexandreLamarre @dbason, from my observation, you can only have 1 audit log file per node. 

To maintain "compatibility" with the rest of the CRD, I decided to use `LogPath`. However, I don't know if we should ask for the complete absolute path (`/path/audit.json`) instead of using the path + fileglob pattern (`/path/*.json`). What do you think?

alexandreLamarre: I can see it both ways, `fileglob` theoretically allows the user to come in with less working knowledge of Kubernetes audit logs, but comes with the risk of overmatching if improperly configured

pkg/resources/collector/templates.go
jaehnri: Changed to `auditlog`.

apis/logging/v1beta1/collector_config_types.go
jaehnri: Following the Rancher Prime directive, I changed it to use something similar to what the [rancher-logging chart does](https://www.suse.com/support/kb/doc/?id=000021022), using `auditFilename` and `pathPrefix`.

dbason (approved):

alexandreLamarre (approved):

PR #1667, opened 2023-08-18
Alerting message fix

alexandreLamarre:
- correctly apply filter funcs in list/withStatus alerting api
- Test endpoint improvements
- Set correct enpoints in cluster drivers
- some test flake fixes & oddball datarace fixes


alexandreLamarre:
Note to self : requires changing the k8s cluster driver peers for the query client to work in a cluster

plugins/alerting/pkg/alerting/alarms/v1/server.go
```diff
@@ -469,22 +473,31 @@ func (a *AlarmServerComponent) Timeline(ctx context.Context, req *alertingv1.Tim
...
 						a.logger.Errorf("failed to get active windows from agent incident tracker : %s", err)
 						return
 					}
-					yieldedValues <- lo.Tuple2[string, *alertingv1.ActiveWindows]{A: cond.Id, B: &alertingv1.ActiveWindows{
-						Windows: activeWindows,
-					}}
+					for _, w := range activeWindows {
+						w.Ref = ref
+					}
+					yieldedValues <- lo.Tuple2[*alertingv1.ConditionReference, *alertingv1.ActiveWindows]{
```
kralicky, line 496:
```suggestion
					yieldedValues <- lo.T2(ref, &alertingv1.ActiveWindows{
						Windows: activeWindows,
					})
```

pkg/apis/alerting/v1/validate.go
```diff
@@ -591,9 +591,12 @@ func (l *ListAlarmMessageRequest) Sanitize() {
 }
 
 func (l *ListAlarmMessageRequest) Validate() error {
-	if l.ConditionId.Id == "" {
+	if l.ConditionId == nil {
 		return validation.Error("field conditionId must be set")
 	}
+	if l.ConditionId.Id == "" {
+		return validation.Error("field conditionId.id must be set")
+	}
```
kralicky, line 599:
```suggestion
	if l.GetConditionId() == "" {
		return validation.Errorf("%w: %s", validation.ErrMissingRequiredField. "conditionId")
	}
```

plugins/alerting/pkg/alerting/endpoints/v1/server.go
```diff
@@ -235,6 +240,36 @@ func (e *EndpointServerComponent) TestAlertEndpoint(ctx context.Context, req *al
 		return nil, err
 	}
 	go func() { // create, trigger, delete
+		start := time.Now()
```
kralicky, line 243: Perhaps this function should just block until ctx is canceled to avoid arbitrary delays

plugins/alerting/pkg/alerting/admin.go
```diff
@@ -218,7 +219,9 @@ func (p *Plugin) Info(ctx context.Context, _ *emptypb.Empty) (*alertops.Componen
 	if err != nil {
 		return nil, status.Error(codes.Unavailable, err.Error())
 	}
+	p.syncController.hashMu.Lock()
```
kralicky, line 222: Not a good idea to use the mutex from here, suggest doing something like `p.syncController.GetHash()` so that the sync controller can manage its own mutex; or, move the mutex into `Plugin`.

plugins/alerting/pkg/node_backend/backend.go
```diff
@@ -84,10 +84,25 @@ var (
...
+var (
+	fallbackDefaultNodeSpec = &node.AlertingCapabilitySpec{
+		RuleDiscovery: &node.RuleDiscoverySpec{
+			Enabled: true,
+		},
+	}
+	specMu sync.RWMutex
+)
+
+func GetFallbackSpec() *node.AlertingCapabilitySpec {
```
kralicky, line 96: The idiomatic way to do this is using a separate file called (for example) `export_test.go` that contains:
```go
package node_backend

var FallbackDefaultNodeSpec = fallbackDefaultNodeSpec
```
Now the variable will be exported only during tests, and unexported otherwise.

kralicky (changes requested):

plugins/alerting/pkg/alerting/alarms/v1/server.go
kralicky: (also below)

pkg/alerting/drivers/routing/routing.go
```diff
@@ -229,6 +229,8 @@ func (o *OpniRouterV1) HasReceivers(routingId string) []string {
 }
 
 func (o *OpniRouterV1) SetDefaultReceiver(cfg config.WebhookConfig) {
+	o.mu.Lock()
```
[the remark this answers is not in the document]
alexandreLamarre: In this case it is not, for example if users want to programmatically set a default receiver, and we must dynamically switch default receivers between embedded server and kubernetes events

plugins/alerting/pkg/alerting/admin.go
alexandreLamarre: this is a stop-gap fix for something else

plugins/alerting/pkg/node_backend/backend.go
alexandreLamarre: I'm not seeing how this idiomatic way can resolve the datarace idiomatically, it leads to split definitions since the mutex has to be used in both the new file and original file

plugins/alerting/pkg/alerting/admin.go
alexandreLamarre: https://github.com/rancher/opni/pull/1672

plugins/alerting/pkg/alerting/endpoints/v1/server.go
alexandreLamarre: the decision was made for the api not to block on a long "send and forget" pipeline, the errors don't return any actionable information for the end user anyways, and the end user may not receive the message until minutes after the end of the non-blocking call anyway

alexandreLamarre: I removed some of the debug logs, and timings

plugins/alerting/pkg/alerting/endpoints/v1/server.go
```diff
@@ -246,15 +281,55 @@ func (e *EndpointServerComponent) TestAlertEndpoint(ctx context.Context, req *al
...
+					Annotations: map[string]string{
+						message.NotificationContentHeader:  "Test notification",
+						message.NotificationContentSummary: "Admin has sent a test notification",
+					},
+					Labels: map[string]string{
+						ns: ephemeralId,
+					},
+				})
+				if err != nil {
+					e.logger.Warnf("failed to trigger alert on test endpoint %s", err)
```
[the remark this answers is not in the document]
alexandreLamarre: intentional that api can return partial failures in rare cases, due to the alertmanager api

pkg/alerting/drivers/routing/routing.go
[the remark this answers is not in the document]
kralicky: Sorry to clarify, was referring to the combination of exported variable + mutex, not the get/set methods necessarily.

plugins/alerting/pkg/node_backend/backend.go
kralicky: Because you don't actually want to modify this unless you're running in a test. And in the test, you can just set it before/after calling the API so it won't race.

plugins/alerting/pkg/alerting/endpoints/v1/server.go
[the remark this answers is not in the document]
kralicky: :+1: suggest putting a comment though since it does look like a mistake at first glance.

pkg/alerting/drivers/routing/routing.go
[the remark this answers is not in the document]
alexandreLamarre: the router is always behind an interface anyways but I can make the field private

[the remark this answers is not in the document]
alexandreLamarre: The router is behind an interface, so accessing fields shouldn't be done anyways, it needs to be public for marshalling purposes

plugins/alerting/pkg/node_backend/backend.go
alexandreLamarre: I'll do what you did for the metrics node_config tests, but this doesn't work with the test structure

plugins/alerting/pkg/alerting/endpoints/v1/server.go
alexandreLamarre: this API is obsolete, it was moved and reworked to the notification server : it was kept in endpoints server to satisfy the existing API contract but forced alerting to do an ugly dependency injection

dbason (approved):

kralicky (approved):

PR #1668, opened 2023-08-18
Replace usages of golang.org/x/exp/{maps,slices} with stdlib packages

kralicky:
The existing version of golang.org/x/exp is kept the same, because it is not compatible with the latest version, and some dependencies still need the old one.


dagger/config/setup.go
```diff
@@ -8,11 +8,12 @@ import (
 	"path/filepath"
 	"strings"
 
+	"slices"
```
jaehnri, line 11: Do we use any convention for import orderings? Is it just stdlib first and then external?

kralicky: I let gopls order the imports however it wants, although it ignores blank spaces between blocks of imports which sometimes requires manually removing them to make it reorder everything.

alexandreLamarre (approved):

Issue #1673, opened 2023-08-23
Alerting / AlertManager generic webhook Endpoint / Receiver


ron1:
Introduce Alerting and AlertManager support for a generic webhook Endpoint / Receiver. AlertManager currently supports a large number of Webhook Receiver integrations as documented [here](https://prometheus.io/docs/operating/integrations/#alertmanager-webhook-receiver). By fully supporting a generic webhook Endpoint, Opni integrates numerous notification mechanisms with just one new Endpoint implemention.

ron1:
Thanks @alexandreLamarre @codyrancher for the quick turnaround implementing this new feature!

PR #1677, opened 2023-08-25
Improve webhook endpoint validation

alexandreLamarre:
Related to #1673

### DIFF
--- a/pkg/alerting/drivers/config/conversion.go
+++ b/pkg/alerting/drivers/config/conversion.go
@@ -1,0 +1,68 @@
+package config
+
+import (
+	"net/url"
+
+	amCfg "github.com/prometheus/alertmanager/config"
+	commoncfg "github.com/prometheus/common/config"
+	alertingv1 "github.com/rancher/opni/pkg/apis/alerting/v1"
+)
+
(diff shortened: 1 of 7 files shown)


pkg/alerting/drivers/config/conversion.go
```diff
@@ -0,0 +1,68 @@
...
+				ClientID:     hc.GetOauth2().GetClientId(),
+				ClientSecret: commoncfg.Secret(hc.GetOauth2().GetClientSecret()),
+				TokenURL:     hc.GetOauth2().GetTokenUrl(),
+			}
+		}
+		if hc.GetProxyUrl() != "" {
+			proxyUrl, err := url.Parse(hc.GetProxyUrl())
+			if err != nil {
+				panic(err)
+				// return validation.Errorf("proxy url must be a valid url : %s", err)
```
kralicky, line 37: intentional?

kralicky (changes requested):

pkg/alerting/drivers/config/conversion.go
alexandreLamarre: yep, I'll remove the comment here, that check is done in `pkg/apis/alerting/v1/validate.go`, so we *should* never get here

kralicky (approved):

PR #1718, opened 2023-09-20
Add storage extension to save filelog receivers' offsets

### LINKED ISSUE #1708: Save collector-agent log file offset in the host
Right now, when an `opni-collector-agent` pod is deleted, it loses the offsets of its file logs causing the sending queue to be full after a restart.

We can use the filelogreceivers' [storage property](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/filelogreceiver) to prevent that.

jaehnri:
Fixes #1708 

The `filestorage` extension allows us to keep track of the tail offset for every log file watched by the `filelog` receivers. This way, whenever an `opni-collector-agent` pod restarts, they can continue reading the log files from where they stopped.

Previously, whenever a crash or restart happened, agents would read from the start again, sending repeated logs and causing an overflow in the aggregator queues:
```
2023-09-20T14:18:42.911Z	error	exporterhelper/queued_retry.go:216	Dropping data because sending_queue is full. Try increasing queue_size.	{"kind": "exporter", "data_type": "logs", "name": "otlp", "dropped_items": 100}
```

Now, on restarts, this is logged to the `opni-collector-agents`:
```
2023-09-20T15:42:14.081Z	info	extensions/extensions.go:31	Starting extensions...
2023-09-20T15:42:14.081Z	info	extensions/extensions.go:34	Extension is starting...	{"kind": "extension", "name": "file_storage"}
2023-09-20T15:42:14.082Z	info	extensions/extensions.go:38	Extension started.	{"kind": "extension", "name": "file_storage"}
2023-09-20T15:42:14.083Z	info	adapter/receiver.go:45	Starting stanza receiver	{"kind": "receiver", "name": "filelog/k8s", "data_type": "logs"}
2023-09-20T15:42:14.084Z	info	fileconsumer/file.go:354	Resuming from previously known offset(s). 'start_at' setting is not applicable.	{"kind": "receiver", "name": "filelog/k8s", "data_type": "logs", "component": "fileconsumer"}
```

Also, I disabled the audit logs by default as `rancher-logging` does.

### DIFF
--- a/pkg/resources/collector/templates.go
+++ b/pkg/resources/collector/templates.go
@@ -232,7 +232,7 @@ processors:
     {{ template "metrics-system-processor" . }}
 extensions:
   file_storage:
-    directory: /etc/otel/filestorage
+    directory: /var/otel/filestorage
     timeout: 1s
 service:
   extensions: [file_storage]
--- a/pkg/resources/collector/templates.go
+++ b/pkg/resources/collector/templates.go
@@ -106,7 +106,7 @@ journald/k3s:
 	templateKubeAuditLogs = template.Must(template.New("kubeauditlogsreceiver").Parse(`
 filelog/kubeauditlogs:
   include: [ {{ . }} ]
-  start_at: beginning
+  storage: file_storage
   include_file_path: false
   include_file_name: false
   operators:
--- a/pkg/resources/collector/templates.go
+++ b/pkg/resources/collector/templates.go
@@ -19,7 +19,7 @@ var (
 filelog/k8s:
   include: [ /var/log/pods/*/*/*.log ]
   exclude: []
-  start_at: beginning
+  storage: file_storage
   include_file_path: true
   include_file_name: false
   operators:
@@ -87,7 +87,7 @@ filelog/k8s:
 	templateLogAgentRKE = `
 filelog/rke:
   include: [ /var/lib/rancher/rke/log/*.log ]
-  start_at: beginning
+  storage: file_storage
   include_file_path: true
   include_file_name: false
   operators:
@@ -157,7 +157,7 @@ journald/rke2:
   directory: {{ . }}
 filelog/rke2:
   include: [ /var/lib/rancher/rke2/agent/logs/kubelet.log ]
-  start_at: beginning
+  storage: file_storage
   include_file_path: true
   include_file_name: false
   operators:
@@ -230,7 +230,12 @@ processors:
       - key: tier
       - key: component
     {{ template "metrics-system-processor" . }}
+extensions:
+  file_storage:
+    directory: /etc/otel/filestorage
+    timeout: 1s
 service:
+  extensions: [file_storage]
   telemetry:
     logs:
       level: {{ .LogLevel }}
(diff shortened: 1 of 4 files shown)


pkg/resources/collector/workloads.go
```diff
@@ -225,6 +232,15 @@ func (r *Reconciler) daemonSet() resources.Resource {
 				},
 			},
 		},
+		{
+			Name: "filestorage-extension",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: otelFileStorageDir,
+					Type: &directoryOrCreate,
```
alexandreLamarre, line 240: This definitely *seems* like the wrong use of of `HostVolume`.
This has the potential to directly affect the underlying host's storage / IO. IMO this is the kind of volatile side effect we would want to avoid especially for agents. Not to mention we want to target low resource environments

We don't have a good way to persist data in the agents at the moment ... Perhaps the way forward is to do something like having the core collector CRD have a field similar to what Opster does where we can specify PVCs to attach and default to something like `EmptyDir` if none are provided

alexandreLamarre: Just to note this immediately brought down my node on a small downstream agent that was having sending queue issues while the gateway was restarting

alexandreLamarre (changes requested):

pkg/resources/collector/workloads.go
jaehnri: Yeah, I agree the current way is potentially dangerous. I'm not familiar with the way Opster does, I'll take a look.

Can you give more details about how the node was brought down? Did it happen on the first installation?

alexandreLamarre: Happened on restart, when updating the image

alexandreLamarre: we synced up with Dan and this just tracks offsets so this should be fine

jaehnri: > Happened on restart when updating the image => the gateway is restarting => gateway is down => agent collector queue filled up

I'm guessing this PR caused the `collector-agents` to restart to pick up the new extension config. During restart, it probably found all the old log files but found no respective offset files. So, the collector-agents interpreted the old and full log files as "new" and began reading them from the start, causing the queues to be full.

This PR literally fixes this bad behavior but also causes it for the very last time. The last dance™.

jaehnri: Not sure if we can do anything to make the upgrade run more smoothly. I'll see if we can use `start_at: end` to prevent that.

alexandreLamarre: I think that's fine

pkg/resources/collector/templates.go
```diff
@@ -230,7 +230,12 @@ processors:
       - key: tier
       - key: component
     {{ template "metrics-system-processor" . }}
+extensions:
+  file_storage:
+    directory: /etc/otel/filestorage
```
alexandreLamarre, line 235: proper convention to use `/var/<something>` for stateful data

jaehnri: Changed to `/var/otel/filestorage`. Could also be `/var/log/otel/filestorage` but I think `/var/log` is more prone to manual deleting, if that makes sense.

alexandreLamarre (approved):

PR #1721, opened 2023-09-20
Refactor roles for backend plugins

dbason:


pkg/apis/core/v1/core.proto
```diff
@@ -122,39 +122,78 @@ message ClusterSelector {
...
+  Role role = 2;
+}
+
+message BackendRoleRequest {
+  CapabilityType capability = 1;
+  core.Reference roleRef = 2;
+}
+
+message RoleList {
+  ReferenceList items = 1;
```
kralicky, line 194: prefer `repeated core.Reference` here otherwise you end up with `roleList.Items.Items`

dbason:
@codyrancher I might need a bit of help with the vue stuff.  The get roles now just returns a list of role ids now rather than a list of role objects.  The idea is that if a user wants to edit a role the api fetches the role when they go into a view role screen - roles are a bit more complex now and don't lend themselves to being viewed in a simple table.

codyrancher:
@dbason 
> @codyrancher I might need a bit of help with the vue stuff. The get roles now just returns a list of role ids now rather than a list of role objects. The idea is that if a user wants to edit a role the api fetches the role when they go into a view role screen - roles are a bit more complex now and don't lend themselves to being viewed in a simple table.

Are you wanting help with creating a new different display? I imagine we still want to display at least a name and maybe group/filter by cluster.

codyrancher (approved):

dbason:
> Are you wanting help with creating a new different display? I imagine we still want to display at least a name and maybe group/filter by cluster.

I think what we have is fine for now, but for v1 we will need to redesign the RBAC stuff.  We can discuss when I'm back from leave

pkg/rbac/middleware.go
```diff
@@ -6,54 +6,76 @@ import (
 
 	"github.com/gin-gonic/gin"
 	corev1 "github.com/rancher/opni/pkg/apis/core/v1"
+	"github.com/rancher/opni/pkg/storage"
+	"go.uber.org/zap"
+	"golang.org/x/exp/slices"
```
alexandreLamarre, line 11: this should probably be `slices`

plugins/metrics/pkg/backend/rbac.go
```diff
@@ -0,0 +1,179 @@
...
+	r.WaitForInit()
+
+	oldRole, err := r.RoleStore.Get(ctx, in.Reference().GetId())
+	if err != nil {
+		return &emptypb.Empty{}, err
+	}
+
+	oldRole.Permissions = in.GetPermissions()
+	oldRole.Metadata = in.GetMetadata()
+	err = r.RoleStore.Put(ctx, oldRole.Reference().GetId(), oldRole)
```
alexandreLamarre, line 107: if the role has immutable fields, we might want to define a helper `apply(old, new)` function or `cmp(old, new) error` function, to prevent improper code changes down the line

alexandreLamarre: Or, perhaps a change to the  `rolestore` interface with an `Update` method that encapsulates the immutable field logic if it is shared between implementations

Issue #1719, opened 2023-09-20
Panic in Alertmanager API


dbason:
```
2023/09/19 02:46:43 http: panic serving 10.42.53.174:45502: write tcp 10.42.53.185:9093->10.42.53.174:45502: write: broken pipe
goroutine 60488 [running]:
net/http.(*conn).serve.func1()
	net/http/server.go:1868 +0xb9
panic({0x4584640?, 0xc0046e3e50?})
	runtime/panic.go:920 +0x270
[github.com/prometheus/alertmanager/api/v2/restapi/operations/alertgroup.(*GetAlertGroupsOK).WriteResponse(0xc003486270](http://github.com/prometheus/alertmanager/api/v2/restapi/operations/alertgroup.(*GetAlertGroupsOK).WriteResponse(0xc003486270), {0x63ecd00, 0xc0029c2000}, {0x63d0f20, 0x5c645b0})
	[github.com/prometheus/alertmanager@v0.25.1-0.20230505130626-263ca5c9438e/api/v2/restapi/operations/alertgroup/get_alert_groups_responses.go:74](http://github.com/prometheus/alertmanager@v0.25.1-0.20230505130626-263ca5c9438e/api/v2/restapi/operations/alertgroup/get_alert_groups_responses.go:74) +0xc5
[github.com/go-openapi/runtime/middleware.(*Context).Respond(0xc001957b30](http://github.com/go-openapi/runtime/middleware.(*Context).Respond(0xc001957b30), {0x63ecd00?, 0xc0029c2000}, 0xc002984e00, {0xc001f7fd70?, 0x1, 0x1}, 0xc002984d00, {0x4504760, 0xc003486270})
	[github.com/go-openapi/runtime@v0.25.0/middleware/context.go:523](http://github.com/go-openapi/runtime@v0.25.0/middleware/context.go:523) +0x6f6
[github.com/prometheus/alertmanager/api/v2/restapi/operations/alertgroup.(*GetAlertGroups).ServeHTTP(0xc001914e88](http://github.com/prometheus/alertmanager/api/v2/restapi/operations/alertgroup.(*GetAlertGroups).ServeHTTP(0xc001914e88), {0x63ecd00, 0xc0029c2000}, 0xc002984e00)
[...]

PR #1759, opened 2023-10-04
Alerting proxy improvements 

alexandreLamarre:
Leverage more robust proxy methods for interacting with the alerting cluster (AlertManager) API

Might resolve https://github.com/rancher/opni/issues/1719


pkg/alerting/client/client.go
```diff
@@ -185,43 +256,41 @@ func (c *Client) ProxyClient() ProxyClient {
 	return c
 }
 
+func (c *Client) scheme() string {
+	return "http"
+}
+
+func (c *Client) target(addr string) string {
+	return c.scheme() + "://" + addr
```
kralicky, line 264: should construct a net.URL here, concatenating url strings can be error-prone

pkg/alerting/client/client_test.go
```diff
@@ -430,67 +426,9 @@ func BuildProxyClientTestSuite(
...
-					_, ok := item.Alert.Labels["manual_test"]
-					return ok
-				})
-
-				Expect(len(proxiedAlerts)).To(BeNumerically(">", 0))
-			})
-
-			It("should be able to list alert groups", func() {
+			It("should construct a proxy target URL", func() {
+				url := cl.ProxyURL()
```
kralicky, line 430: if you intend to keep the panic, this should include something like Expect(cl.ProxyURL).NotTo(Panic())

pkg/alerting/extensions/embedded_server_test.go
```diff
@@ -376,7 +387,9 @@ func BuildEmbeddedServerNotificationTests(
 			By("verifying the alerting cluster has received unique alerts for each unique fingerprint")
 			Eventually(func() error {
 				ags, err := alertingClient.AlertClient().ListAlerts(context.TODO())
```
kralicky, line 389: Use spec context here

kralicky (approved):

PR #1766, opened 2023-10-10
Updating the way we manage the cortex config

codyrancher:
During the first configuring we do the following:
- Fetch the active config
- Merge the first preset in the list of presets with the active config
- Merge the user config

During subsequent configurations we do the following:
- Fetch the active config
- If the user selects a new preset and press the apply button we'll merge the preset over the active config
- Merge the user config

We also now pass the revision that we get from the active config now

image


codyrancher:
While testing this I did run into revision conflict issues on a couple occasions. Unfortunately it would report `no 13 wire type 7`. If you see this I published a pr to resolve here https://github.com/rancher/opni/pull/1794

dbason (approved):

PR #1774, opened 2023-10-16
Stream agent logs remotely

### LINKED ISSUE #1655: Management API to stream agent logs
We should implement a simple management API to allow viewing logs for any connected agent. An API should look something like this:

```proto
package control;
service Log {
  // Streams logs from the agent according to the parameters specified in the request.
  // The stream will be closed when the newest requested log is sent, unless 'follow'
  // is true, in which case the stream will remain open indefinitely and continue to
  // send logs in real time as they are generated.
  rpc StreamLogs(LogStreamRequest) returns (stream StructuredLogRecord);
}

message LogStreamRequest {
  google.protobuf.Timestamp since   = 1;
  google.protobuf.Timestamp until   = 2;
  repeated LogStreamFilter  filters = 3;
  bool                      follow  = 4;
}

message LogStreamFilter {
  // A glob pattern to match against the logger name. For example, "plugin.*"
  // will match all plugin loggers.
  optional string namePattern = 1;
  // Minimum log severity level to include in the stream.
  optional int32 level = 2;
}

// Represents a single structured log message. Roughly analagous to slog.Record.
message StructuredLogRecord {
  google.protobuf.Timestamp time       = 1;
  string                    message    = 2;
  repeated Attr             attributes = 3;
}

// A key-value pair analagous to slog.Attr.
message Attr {
  string              key   = 1;
  google.protobuf.Any value = 2;
}


```

jan-law:
Fixes #1655

Depends on #1642


pkg/logger/logger.go
```diff
@@ -96,181 +80,167 @@ func (o *LoggerOptions) apply(opts ...LoggerOption) {
...
+		f, err := logFs.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0666)
+		if err != nil {
+			panic(err)
+		}
+		// FIXME where to close this file?
+
+		logFileHandler := newProtoHandler(f, ConfigureProtoOptions(options))
+
+		// distribute logs to handlers in parallel
+		return slogmulti.Fanout(handler, logFileHandler)
```
jan-law, line 205: `logFs.OpenFile` sometimes causes a data race when the `proto_handler` writes to the log file with `_, err = h.w.Write(*buf)` and the `seek` call in `OpenFile` [here](https://github.com/spf13/afero/blob/master/memmap.go#L262). If I remove the `os.O_APPEND` flag, not all the logs get written to the file. Any suggestions to handle the file correctly?

stack trace: https://gist.github.com/jan-law/eb20c79a98ba78f1091cf941e748e8ef

plugins/metrics/pkg/agent/drivers/opni_manager_otel/otel_driver.go
```diff
@@ -253,7 +252,7 @@ func init() {
 	drivers.NodeDrivers.Register("opni-manager-otel", func(_ context.Context, opts ...driverutil.Option) (drivers.MetricsNodeDriver, error) {
 		options := OTELNodeDriverOptions{
 			Namespace: os.Getenv("POD_NAMESPACE"),
-			Logger:    logger.NewPluginLogger().Named("metrics").Named("otel"),
+			Logger:    logger.NewPluginLogger().WithGroup("metrics").WithGroup("otel"),
```
jan-law, line 255: Some plugin loggers are created during `init()`, before I can call the `identityClient` to get the agent ID corresponding to the log file. Is there another way to initialize or pass in these loggers?

jan-law (commented):
Added two questions about writing plugin loggers to log files. Once those are resolved, this PR is ready for review

pkg/logger/logger.go
kralicky: Appends are only atomic under certain conditions. (see the section `PIPE_BUF` in `man 7 pipe` for details). You might need to change the O_NONBLOCK flag and/or write in chunks of 4096 bytes

jan-law: From the man page, it looks like writes of > 4096 bytes are not guaranteed to be atomic regardless if O_NONBLOCK is set. I ended up writing in chunks of 4096 bytes.

plugins/metrics/pkg/agent/drivers/opni_manager_otel/otel_driver.go
jan-law: Added a `pluginWriter` that gets initialized after the agent id is available. Let me know what you think!

jan-law:
When running the `opni debug agent-logs` in my k3s cluster, I get an `Unrecognized remote plugin message`  for one of the log messages. Do you know where this log message is created?

Tried searching here and couldn't find it: https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-go-contrib+%22Tracer+created%22&type=code
```
2023 Oct 18 20:51:14 ERROR agent.pluginloader plugins/loader.go:156 failed to load plugin plugin=github.com/rancher/opni/plugins/logging err=Unrecognized remote plugin message: 2023 Oct 18 20:51:14 DEBUG global/internal_logging.go:53 Tracer created logger=tracing name=go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc version=0.42.0 schemaURL=""
This usually means
  the plugin was not compiled for this architecture,
  the plugin is missing dynamic-link libraries necessary to run,
  the plugin is not executable by this process due to file permissions, or
  the plugin failed to negotiate the initial go-plugin protocol handshake

Additional notes about plugin:
  Path: /var/lib/opni-agent/plugins/plugin_logging
  Mode: -rwxr-xr-x
  Owner: 0 [root] (current: 0 [root])
  Group: 0 [root] (current: 0 [root])
  ELF architecture: EM_X86_64 (current architecture: amd64)
```

kralicky:
@jan-law that message is generated by the go-plugin library when writing to a logger that isn't set up to forward its log messages to the host process.

jan-law:
> that message is generated by the go-plugin library when writing to a logger that isn't set up to forward its log messages to the host process.

Update - it seems that go-plugin expects logs to be sent to stderr instead of stdout. This issue affects the slog branch too. Making a fix for it now.

pkg/logger/logger.go
```diff
@@ -24,18 +25,23 @@ var (
...
+	logFs             afero.Fs
+	logFileName       = "opni-logs"
 	DefaultTimeFormat = "2006 Jan 02 15:04:05"
 	errKey            = "err"
 )
 
 var logSampler = &sampler{}
 
+func init() {
+	logFs = afero.NewMemMapFs()
```
jan-law, line 47: The plugin logs do not appear in the `debug agent-logs` output. Do the plugins share the same `logFs` here if they are different processes?

kralicky: No, they will be separate. The plugins should be sending their logs to the host process though, maybe that needs some refactoring.

jan-law: Although I wasn't able to stream the logs over `SyncStderr`, the `Logger` in the `ClientConfig` can forward the log messages to a writer, where I can log them with a logger on the host side.
```
Logger: hclog.New(&hclog.LoggerOptions{
			Level:  hclog.Debug,
			Output: logger.PluginFileWriter,
		}),
```
Go-plugin can't directly forward protobuf message bytes unfortunately, as it's hardcoded to cast all non-hclog output to a `string`: https://github.com/hashicorp/go-plugin/blob/017b758bf4d495212a55db3de61b2d95ab104e53/client.go#L1155

jan-law: The forwarded logs look like this. What do you think of the formatting?
image

jan-law:
Depends on https://github.com/alexandreLamarre/go-plugin/pull/2

pkg/logger/logger.go
jan-law: Log streaming is working with the forked go-plugin patch in my k3s cluster. I replaced the hclog forwarding logger with `io.Copy(c.config.Stderr, r)`. I'm working on fixing the test env output because the plugins are logging the gprc bytes to stderr

jan-law:
Setting this PR back to draft mode while I work on the plugin loggers for the standalone test env

pkg/opni/commands/debug.go
```diff
@@ -462,6 +467,85 @@ func BuildDebugImportAgentCmd() *cobra.Command {
...
+					attrs.WriteString("=")
+					attrs.WriteString(attr.Value)
+					attrs.WriteString(" ")
+				}
+				fmt.Println(log.Time.AsTime().Format(logger.DefaultTimeFormat), log.Level, log.Name, log.Source, log.Message, attrs.String())
+			}
+		},
+	}
+
+	cmd.Flags().StringVar(&since, "since", "1 day ago", "Start time")
```
kralicky, line 540: Can you add flag completions for some of these? e.g. 
```go
	cmd.RegisterFlagCompletionFunc("level", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"debug", "info", "warn", "error"}, cobra.ShellCompDirectiveDefault
	})
	cmd.RegisterFlagCompletionFunc("output", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
		return []string{"text", ...}, cobra.ShellCompDirectiveDefault
	})
```

pkg/opni/commands/debug.go
```diff
@@ -462,6 +467,85 @@ func BuildDebugImportAgentCmd() *cobra.Command {
...
+				}
+
+				var attrs strings.Builder
+				for _, attr := range log.Attributes {
+					attrs.WriteString(attr.Key)
+					attrs.WriteString("=")
+					attrs.WriteString(attr.Value)
+					attrs.WriteString(" ")
+				}
+				fmt.Println(log.Time.AsTime().Format(logger.DefaultTimeFormat), log.Level, log.Name, log.Source, log.Message, attrs.String())
```
kralicky, line 535: Should this use `slog.Log(...)`?

jan-law: Thanks, this reminded me I need to implement `output=json` - I would use `slog.Log` there, eg
```
handlerOptions := &slog.HandlerOptions{
		Level: logger.ParseLevel(level), 
		AddSource: false, 
		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
		if a.Key == slog.TimeKey && len(groups) == 0 {
			return slog.Attr{}
		}
		return a
		},
	}

if output == "json" {
      slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stderr, handlerOptions)))
}
...

slog.Log(context.TODO(), logger.ParseLevel(log.GetLevel()), log.GetMessage(), attrs...)

```

For `output=text` though, if I use `slog.SetDefault(logger.New())`, I'm unable to overwrite the `<time user called opni debug> ` timestamp while preserving the log formatting. I could write a handler that assumes there will be a `time` attribute in every call to slog.Log and replace the original timestamp, except I don't want to rely on that assumption.

Here are some alternatives to `fmt.Println`. What do you think?

Option 1: preserve tinted colors but leave the double timestamp
```
slog.SetDefault(logger.New(logger.WithDisableCaller()))
```
image

Option 2: use default slog TextHandler, and remove the debug cli time/source
```
slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: logger.ParseLevel(level), AddSource: false, ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
					if a.Key == slog.TimeKey && len(groups) == 0 {
						return slog.Attr{}
					}
					return a
				},
				})))
```
image

PR #1786, opened 2023-10-20
Rebase permissions changes for gateway scaling

dbason:


kralicky (approved):

pkg/rbac/middleware.go
```diff
@@ -4,56 +4,79 @@ import (
 	"context"
 	"net/http"
 
+	"log/slog"
+
 	"github.com/gin-gonic/gin"
 	corev1 "github.com/rancher/opni/pkg/apis/core/v1"
+	"github.com/rancher/opni/pkg/storage"
+	"golang.org/x/exp/slices"
```
alexandreLamarre, line 12: should probably be `slices`

plugins/metrics/pkg/gateway/services/rbac.go
```diff
@@ -0,0 +1,257 @@
...
+			innerErr = err
+			return
+		}
+		revision, err := strconv.ParseInt(in.GetMetadata().GetResourceVersion(), 10, 64)
+		if err != nil {
+			innerErr = err
+			return
+		}
+		oldRole.Permissions = in.GetPermissions()
+		innerErr = r.rolesStore.Put(ctx, oldRole.Reference().GetId(), oldRole, storage.WithRevision(revision))
```
alexandreLamarre, line 147: if the role has immutable fields, we could consider defining a helper `apply(old, new)` function or `cmp(old, new)` error function, to prevent improper code changes down the line

Or, perhaps a change to the `rolestore` interface with an `Update` method that encapsulates the immutable field logic if it is shared between implementations

plugins/metrics/pkg/gateway/services/rbac.go
```diff
@@ -0,0 +1,257 @@
+package services
```
alexandreLamarre, line 1: In general I would discourage using `system.NewLock` here since these are meant to be long lived locks that are intended to live for the duration of the process -- the only way such a lock is guaranteed to unlock is if the client holding it is unreachable, which is probably not desireable for transactions (or unless the client making requests for the lock only lives for the duration of the API call, that would be valid, but not possible if using the `system.NewLock` abstraction)

Also, in particular the etcd implementation in `topic/scaling` of the etcd Lock will not lock between the same host as it reuses the underlying client connection, so I don't think this would achieve the effect desired here. Somewhat ironically, adding a sync.Mutex would fix that

I would suggest using revisions and embedding them in the interface, since the underlying implementation is the KV store anyways

alexandreLamarre (commented):
also a reminder to remove commented out code unless there is a reason to keep it as reference

plugins/metrics/pkg/gateway/services/rbac.go
dbason: This is a direct conversion of the existing logic.  I think improvements like this could be made separately.

alexandreLamarre (approved):

PR #1793, opened 2023-10-25
Cert & alerting fixes

alexandreLamarre:
- Fixed a change lost in one of the rebases; cluster drivers are only responsible for finding the alerting cluster peers, not initializing clients
- Fixed a bug where an initial install of the alerting cluster would always result in a TLS error
- Fixed a missing edge case where prometheus rules not being present in the metrics backend resulted in an `Ok` status instead of `Pending`
- Fixed a UI bug that made it impossible to delete alert conditions
- Improved the detection of the existence of rule objects in cortex from the alerting plugin

### DIFF
--- a/plugins/alerting/pkg/alerting/drivers/alerting_manager/options.go
+++ b/plugins/alerting/pkg/alerting/drivers/alerting_manager/options.go
@@ -1,10 +1,12 @@
 package alerting_manager
 
 import (
+	"crypto/tls"
+	"log/slog"
+
 	alertingClient "github.com/rancher/opni/pkg/alerting/client"
 	"github.com/rancher/opni/pkg/alerting/shared"
 	"k8s.io/apimachinery/pkg/types"
-	"log/slog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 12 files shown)


alexandreLamarre:
This has a breaking change with evaluating Prometheus based alert status. All prometheus rules will show up as `Pending` -- uninstalling and reinstalling the alerting cluster will properly index the Prometheus based alerts

kralicky (dismissed):

kralicky (approved):

plugins/alerting/pkg/alerting/drivers/alerting_manager/options.go
```diff
@@ -16,4 +18,5 @@ type AlertingDriverOptions struct {
 	InternalRoutingKey string                               `option:"internalRoutingKey"`
 	AlertingOptions    *shared.AlertingClusterOptions       `option:"alertingOptions"`
 	Subscribers        []chan alertingClient.AlertingClient `option:"subscribers"`
+	TlsConfig          *tls.Config                          `option:"tlsConfig"`
```
jaehnri, line 21: I **really** don't wanna be that guy but acronyms should be in consistent case -> TLSConfig

jaehnri (approved):

PR #1799, opened 2023-10-30
Add api gateway proxy and authc/authz

dbason:


pkg/auth/middleware/middleware.go
```diff
@@ -0,0 +1,121 @@
...
+			c.Set(proxy.SubjectKey, "OPNI_admin")
+			return
+		}
+		if errors.Is(err, bcrypt.ErrMismatchedHashAndPassword) {
+			c.Header(authenticateHeader, authRealm)
+			c.AbortWithStatus(http.StatusUnauthorized)
+			m.Logger.Warn("auth failed for admin")
+			return
+		}
+		m.Logger.With("error", err.Error()).Error("password verification failed")
```
alexandreLamarre, line 97: (we have a shortcut `logger.Err(err)` for this in pkg/logger)

pkg/dashboard/dashboard.go
```diff
@@ -150,11 +179,71 @@ func (ws *Server) ListenAndServe(ctx context.Context) error {
...
+			Store:  ws.ds.StorageBackend(),
+			Logger: log,
+			Client: p,
+		})
+		if err != nil {
+			log.With(
+				logger.Err(err),
+			).Error("failed to create plugin router")
+			return
+		}
```
alexandreLamarre, line 200: should this be a panic?

pkg/proxy/backend/backend.go
```diff
@@ -0,0 +1,67 @@
...
+	RewriteProxyRequest(string, *corev1.ReferenceList) (func(*httputil.ProxyRequest), error)
+}
+
+type implBackend struct {
+	backendURL   string
+	pluginClient proxyv1.RegisterProxyClient
+	logger       *slog.Logger
+}
+
+func NewBackend(logger *slog.Logger, client proxyv1.RegisterProxyClient) (Backend, error) {
```
alexandreLamarre, line 25: If this backend is always intended to be used inside a plugin, the constructor and struct should accept a context so that it can inherit the plugin's context

pkg/dashboard/oidc.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/coreos/go-oidc/v3/oidc"
+	"github.com/gin-gonic/gin"
+	"github.com/rancher/opni/pkg/storage"
+	"golang.org/x/oauth2"
+)
+
+// AuthDataSource provides a way to obtain data which the auth
+// middleware and proxies need.
+type AuthDataSource interface {
+	StorageBackend() storage.Backend
```
alexandreLamarre, line 16: does this need to encapsulate all the `storage.Backend` methods? if not, it might be better to have a more granular interface here (for separation of concerns and future proofing)

alexandreLamarre (approved):

pkg/proxy/backend/backend.go
dbason: The backend works the other way around.  It takes in a plugin client which it uses to call down to the plugin to construct the auth headers to inject in the request to the backend.

pkg/dashboard/oidc.go
dbason: We need multiple interfaces (KeyValueBroker and RoleBindingStore) so to keep the types matched up correctly we need to use the top level interface.

pkg/dashboard/dashboard.go
dbason: I don't think so.  This follows the standard pattern where if we're unable to load a plugin we make the best effort to continue without taking down the whole system.

internal/cmd/testenv/main.go
```diff
@@ -244,7 +245,9 @@ func main() {
 			if !enableGateway {
 				return
 			}
-			opts := []dashboard.ServerOption{}
+			opts := []dashboard.ServerOption{
+				dashboard.WithLocaalAuthenticator(&localauth.TestLocalAuthenticator{}),
```
kralicky, line 249: typo

pkg/apis/management/v1/management.proto
```diff
@@ -135,6 +146,30 @@ service Management {
       get: "/rbac/backend/{name}/roles"
     };
   }
+  rpc AddAdminRoleBinding(core.Reference) returns (google.protobuf.Empty) {
```
kralicky, line 149: Making these separate apis entirely seems like the wrong move, since when we undo this it will necessarily be a breaking change. Is there a better way?

pkg/apis/proxy/v1/proxy.proto
kralicky, line 1: formatting

pkg/test/localauth/localauth.go
```diff
@@ -0,0 +1,21 @@
...
+
+type TestLocalAuthenticator struct{}
+
+func (a *TestLocalAuthenticator) GenerateAdminPassword(_ context.Context) (string, error) {
+	return testPassword, nil
+}
+
+func (a *TestLocalAuthenticator) ComparePassword(_ context.Context, pw []byte) error {
+	if string(pw) != testPassword {
+		return errors.New("mismatched password")
```
kralicky, line 18: Wouldn't this need to be bcrypt.ErrMismatchedHashAndPassword? Alternatively, we could have our own error for this.

pkg/util/oidc/oidc.go
```diff
@@ -0,0 +1,48 @@
...
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"strings"
+
+	"golang.org/x/oauth2"
+)
+
+func SubjectFromClaims(lg *slog.Logger, token *oauth2.Token, claimName string) *string {
+	rawToken, ok := token.Extra("id_token").(string)
```
kralicky, line 14: Don't parse this yourself; use https://github.com/rancher/opni/blob/131a150882dcb208b2eed500cb1ba570e9678d87/pkg/auth/openid/tokens.go#L38-L41

kralicky (changes requested):

pkg/apis/management/v1/management.proto
dbason: I don't think it will be breaking.  Once we create more granular permissions this would simply use a different RPC

kralicky (approved):

PR #1822, opened 2023-11-02
prevent AiOps modeltraining application deadlocks

### LINKED ISSUE #1815: Workload AIOps "deadlock" issue / Cannot uninstall AiOps
I had previously configured the deployment watchlist (~26 pods accross 3 clusters), but its not showing up here

Deployment watch list issue : 

https://github.com/rancher/opni/assets/56211853/6f07f690-b256-451b-aa79-58d71ca30d4a

Results in not being able to uninstall AiOps at all

alexandreLamarre:
Closes #1815 

Here the application deadlock refers to the fact that:

A model training job was scheduled but potentially failed, and cannot be deleted unless manually deleting the stateful information directly from the KV store. And it then becomes impossible to delete, update the watchlist as well as uninstall AiOps.

Fixed by:

- Adding an option to purge stateful model training info when uninstalling

- Fixes model training request flow logic:
  - If the request is never received by the training controller (e.g. because the Pod is pending, which happened > 50% of the time on my cluster), we would have both parameter and status data, causing a deadlock in the application without training a model


plugins/aiops/pkg/gateway/admin.go
```diff
@@ -338,14 +341,38 @@ func (s *AIOpsPlugin) DeleteAISettings(ctx context.Context, _ *emptypb.Empty) (*
...
-	return &emptypb.Empty{}, nil
+	deleteEg.Go(func() error {
+		if err := s.deleteAIOpsResources(ctx); err != nil {
+			s.Logger.Error(fmt.Sprintf("failed to delete AiOps resources : %s", err))
+			return err
+		}
+		return nil
+	})
+	deleteEg.Wait()
+	return &emptypb.Empty{}, deleteEg.Error()
```
dbason, line 375: For the UI flow this should be serialized with the KV delete occurring first (as discussed)

dbason (changes requested):

dbason (approved):

AmartC (approved):
LGTM but make sure to pass the build before merging.

PR #1828, opened 2023-11-06
Fix k3s processing

dbason:


pkg/resources/preprocessor/workloads.go
```diff
@@ -70,8 +70,13 @@ processors:
     log_statements:
     - context: log
       statements:
-      - set(attributes["COMM"], attributes["_COMM"])
-      - delete_matching_keys(attributes, "^_.*")
+      - merge_maps(cache, body, "upsert") where not IsString(body) and attributes["filename"] == nil
```
jaehnri, line 73: How can the body not be a string? What else can it be?

jaehnri (approved):

pkg/resources/preprocessor/workloads.go
jaehnri: Ah, it can be a map, right?

alexandreLamarre (approved):
LGTM and validated with k3s 1.26.8

pkg/resources/preprocessor/workloads.go
dbason: In the case of journald logs it is a map.  That is specifically the case we want to handle.

PR #1833, opened 2023-11-06
Validation

kralicky:
This PR contains new validation code (initially just for metrics apis) using https://github.com/kralicky/protovalidate-go (forked from https://github.com/bufbuild/protovalidate-go) and https://github.com/kralicky/protovalidate (forked from https://github.com/bufbuild/protovalidate). Additionally, there are improvements to the cortex config proto generator tools to allow preserving some modifications to the generated code, and ensuring updates preserve backwards compatibility.

The cel repo is here https://github.com/google/cel-go but it's not particularly useful. However there are a _ton_ of examples here: https://github.com/bufbuild/protovalidate/tree/main/examples


pkg/opni/root.go
```diff
@@ -56,7 +56,7 @@ func Execute() {
 		os.Exit(1)
 	}()
 	if err := rootCmd.ExecuteContext(ctx); err != nil {
-		rootCmd.PrintErrln(err.Error())
+		// rootCmd.PrintErrln(err.Error())
```
alexandreLamarre, line 59: commented out code

pkg/plugins/driverutil/config.go
```diff
@@ -311,6 +318,18 @@ func (ct *DefaultingConfigTracker[T]) History(ctx context.Context, target Target
 	return revisions, nil
 }
 
+func (ct *DefaultingConfigTracker[T]) runValidation(conf T) (*protovalidate.ValidationError, error) {
+	err := ct.validator.Validate(conf)
+	var valErr *protovalidate.ValidationError
+	if err != nil {
+		if !errors.As(err, &valErr) {
+			// invalid validation rules, etc.
+			return nil, status.Error(codes.Internal, err.Error())
```
alexandreLamarre, line 327: In what cases would we encounter invalid validation rules? Does protobuf generation pass if we have invalid generation rules?

pkg/util/generic.go
```diff
@@ -101,6 +101,20 @@ func FieldIndexByName[T proto.Message](name string) int {
 	return -1
 }
 
+type Collection[T any] interface {
+	Len() int
+	Get(i int) T
+}
+
+func Collect[T any, C Collection[T]](c C) []T {
```
alexandreLamarre, line 109: `Collect` is a little ambiguous

alexandreLamarre (approved):

pkg/plugins/driverutil/config.go
kralicky: Currently, invalid cel expressions are not a compiler error, but the upstream typed builtin options are. They are validated, I just haven't implemented surfacing the warnings to the diagnostic handler yet so that they can be handled like actual compiler errors. 
For now, invalid expressions won't have the embedded syntax highlighting, like this:
image
image

and the cel errors are logged in the `Protobuf Language Server` vscode output:
```
ERROR: <input>:1:50: Syntax error: token recognition error at: '= '
 | (this.type == 0 && has(this.etcd)) || (this.type = 1 && has(this.jetstream))
 | .................................................^
```

pkg/util/generic.go
kralicky: I don't think this was actually used, so I can remove it for now. It's specifically for these guys 
image

alexandreLamarre (approved):

dbason (approved):

PR #1842, opened 2023-11-13
Update OpenSearch from 2.8.0 to 2.11.0

jaehnri:


jaehnri:
Won't work for now, 2.11.0 is not supported by the opensearch-k8s-operator.

PR #1872, opened 2023-11-22
Revert "bump pulumi sdk; increase node root volume size"

kralicky:
This reverts commit 5cf0f267abd0ee8ca7b4d1b97f4b40e99077f251.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -169,7 +169,7 @@ require (
 
 require (
 	github.com/go-logr/zapr v1.2.4 // indirect
-	go.uber.org/zap v1.25.0 // indirect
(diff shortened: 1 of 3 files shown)


jaehnri (approved):
While digging it,  I found we can maintain most of the dependency updates. What breaks is upgrading `github.com/pulumi/pulumi/sdk/v3` to `>= v3.88.0`, triggering `golang.org/x/mod v0.13.0`.

But I'm fine with reverting it aswell.

kralicky:
Yeah I noticed that as well, but reverting it will make the history cleaner -- we'll be un-reverting it at some point anyway.

PR #1879, opened 2023-11-28
Fix slog handler edge cases

jan-law:
- logging nil pointer to `encoding.TextMarshaler`
- updates tests to reflect decision to enforce timestamps on all logs
- fixes typo for expected output when dropping an attribute

### DIFF
--- a/pkg/logger/color_handler.go
+++ b/pkg/logger/color_handler.go
@@ -348,7 +348,7 @@ func (h *colorHandler) appendValue(buf *buffer, v slog.Value, shouldQuote bool) 
 		case *slog.Source:
 			h.appendSource(buf, cv.File, cv.Line)
 		default:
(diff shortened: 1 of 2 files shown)


pkg/logger/color_handler.go
```diff
@@ -335,6 +336,10 @@ func (h *colorHandler) appendValue(buf *buffer, v slog.Value, shouldQuote bool)
 		case slog.Level:
 			h.appendLevel(buf, cv)
 		case encoding.TextMarshaler:
+			if reflect.ValueOf(cv).IsNil() {
+				h.appendString(buf, "", shouldQuote)
+				break
+			}
```
kralicky, line 342: There's also line below here that it won't let me comment on
`h.appendString(buf, fmt.Sprint(v.Any()), shouldQuote)`
instead of calling v.Any() again can you use `cv`?

